PR #2, opened 2022-06-25
Compile error looking for WiFiData.h

mascencerro:
When compiling 'hello world' with beken-72xx-arduino framework, compile would fail with the following:
`.platformio/platforms/libretuya/arduino/libretuya/api/WiFi/WiFi.h:32:22: fatal error: WiFiData.h: No such file or directory`

File was not present in repository.

Copied WiFiData.h from arduino/realtek-ambz/libraries/WiFi/WiFiData.h
Changed include for WiFiData.h from  <> to ""

Compile now completes.

### DIFF
--- a/arduino/libretuya/api/WiFi/WiFi.h
+++ b/arduino/libretuya/api/WiFi/WiFi.h
@@ -29,7 +29,7 @@
 
 #include "WiFiType.h"
 // family's data structure
-#include <WiFiData.h>
+#include "WiFiData.h"
 
 class WiFiClass {
   public:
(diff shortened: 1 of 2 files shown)


kuba2k2:
As per the [Implementation status](https://kuba2k2.github.io/libretuya/docs/implementation-status/) page, BK72xx doesn't support WiFi yet. It should be available in the coming days, hopefully even today.

If this error occurred without including WiFi.h at all (in your hello world program) let me know so I'll look into it.

Note that WiFiData.h is a family specific file, so you can't just put it into the common core. Refer to [Project structure](https://kuba2k2.github.io/libretuya/docs/project-structure/) for more details.

kuba2k2:
Fixed with 79ccd47f71c75f263ad449431077ab250a782ab6.

Issue #27, opened 2022-10-03
[esphome] Dallas, rotary_encoder, ht711 and dht not working


dbuezas:
## Compilation error:
```
undefined reference to `esphome::InterruptLock::InterruptLock()'
```

## Due to it not being defined for LibreTuya
<img width="538" alt="image" src="https://user-images.githubusercontent.com/777196/193578899-8be94d89-bcc1-422a-b040-4a5ed06ddf10.png">

## Solution
`#elif defined(USE_ESP32) || defined(USE_LIBRETUYA)`

dbuezas:
Dallas (and others probably too) use micros() inside interruptLock blocks, which is also broken.

kuba2k2:
Fixed in #28.

PR #28, opened 2022-10-03
[beken-72xx] Fix micros() usage with disabled interrupts

dbuezas:
Relates to #27.
Workaround described in comments:

[micros can go backwards]
```cpp
/*
This happens if micros is called from within a interruptLock block (interrupts disabled).
In this case, if the tick counter rolls over, millis() won't be updated, and micros will
be lag by 1 rollover.
The workaround only works as long as micros() calls happen within 2ms of eachother.
WARNING: if interrupts are disabled for more than 2ms, micros() and millis() will temporarily get out of
sync.
*/
```

### DIFF
--- a/arduino/beken-72xx/cores/arduino/wiring.c
+++ b/arduino/beken-72xx/cores/arduino/wiring.c
@@ -66,16 +66,16 @@ unsigned long micros() {
 #endif
 
 #if LT_MICROS_HIGH_RES
-	static uint32_t lastMillis		  = 0;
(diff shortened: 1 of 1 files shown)


kuba2k2 (approved):

dbuezas:
I'm happy with the code now. I'll let it running for the night just in case and report back.

The dallas temperature sensor is working very well and the board seems stable.
image

kuba2k2 (approved):

PR #29, opened 2022-10-07
I think this'll fix the links

blakadder:


blakadder:
nvm i need to build it locally to see what went wrong

PR #68, opened 2023-01-28
[beken-72xx] Fix for flicker when PWM lights turn on/off on bk7231n

mattshepcar:
Some of my BK7231N RGBWW bulbs were occasionally flickering when they turned on/off with libretuya-esphome.  This was more noticeable at higher PWM frequencies.  I discovered that when the pwm dutycycle value is set to 0 then the output goes high.  This doesn't seem to affect my BK7231T based bulbs.


kuba2k2:
Thanks for this! The commit seems to change behavior on T as well (no duty cycle setting), why does that line need to be removed?

mattshepcar:
Sorry, I didn't explain that bit: even momentarily setting the duty cycle to 0 before turning PWM mode off was causing a flicker.

kuba2k2 (approved):

Issue #17, opened 2022-09-08
CBx boards - missing


kuba2k2:

ianchi:
What is missing to use CBU board?
I think it should be supported as it uses BK7231N chip.
How can I help to include support?

kuba2k2:
The board diagrams are missing (the ones that boardgen uses to write docs/pinouts). They're not that hard to create if you know how to, but it's not really documented anywhere. Alternatively, it could probably work without the diagram (i.e. only the pin mapping) but I haven't tried it yet.

PR #106, opened 2023-03-26
[boards] Add Tuya CBU board

ianchi:
This is a draft to include Tuya CBU board definition. (see #17)

A new template needs to be added to `boardgen`. It is included in this PR as reference, I could send a separate PR to `boardgen`.
It is not a perfect drawing, but a good sketch, reusing the current pins shapes and antenna, although dimensions are not exact for the actual cbu board.

Not sure how to generate all the `.h`, `.cpp` and update the rest of the stuff

pinout_cbu


kuba2k2:
Thanks, this looks pretty good, minus the few drawing problems. I'll probably review it after the structure refactor, though (or: merge it directly into the refactor branch).

kuba2k2:
Hi
It took a little bit longer than expected. I ended up refactoring the shapes and boardgen internals, which made it very easy to rewrite the CBU template as well (to match the exact dimensions and shape of the board). I kept the pinouts and other JSONs you prepared, however.
Building boards should be much easier now, as everything can be edited in boardgen GUI (which is a ltchiptool plugin).
Thanks for the PR!

PR #111, opened 2023-04-24
[beken-72xx] Implement WiFi powersave

Xmister:


kuba2k2:
What does the powersave function accomplish? Less power consumption without disconnection, or what? (I don't really understand the whole powersave concept)

kuba2k2 (approved):

rodneygp:
Just to make sure I understand this correct: are BK-based installations have power save on by default now? Or how do I configure it to be on?

Cossid:
LibreTiny is a platform, so whatever uses it as an implementation must set it.  If you're implementing directly, you must call `WiFiClass::setSleep(bool enable)`, if you're using ESPHome, you would need to set a [wifi power save mode](https://esphome.io/components/wifi#wifi-power-save-mode).  LibreTiny only supports on or off, so `LIGHT` and `HIGH` will give identical results.

PR #147, opened 2023-07-05
[realtek-ambz] Improve flashing guide, add test pads for WR2 board

Sonic-Amiga:
Known USB adapter compatibility issues. Emphasize, so that the paragraph is easier to notice.


Sonic-Amiga:
@kuba2k2 Hello, this is me again. I also want to improve a page on WR2 with serial port pin images and photos, but cannot find where it's located. Searching for an .svg file or an .md file with text yields nothing. Can you point me at a right location ?

kuba2k2:
The pinout pages are auto-generated by boardgen. Adding pins needs some work to make the graphic look good, and currently I'm the only known person who can do that 😂 You're welcome to try, nevertheless

Sonic-Amiga:
Not familiar with the tool, okay, but anyways, where are source files then? And where is all the accompanying text? I don't see it either

kuba2k2:
The tool is here:
https://github.com/kuba2k2/boardgen
and the source files are here:
https://github.com/kuba2k2/libretiny/tree/master/boards

As it's auto-generated, there are no "source text files" or images.

If you install https://github.com/libretuya/ltchiptool using Python (the .exe won't work in this case) and install boardgen from PyPI, you'll have a GUI editor for the pinout diagrams.

Sonic-Amiga:
Just finished my first try. After some digging in JSONs and trial and error, i think i was able to make it. Please take a look. I didn't know about the GUI editor, all done by hands.

kuba2k2:
See cb2s-test.json for an example how should test pads be defined. Doing it in the template will but them in all boards, not only wr2.

Same goes for the flashing notice - it's not always the case that it's exposed on the back side (i.e. WR3). Also, the info is already present in the flashing guide, which is linked in the document.

Sonic-Amiga:
Modified.

Sonic-Amiga:
@kuba2k2 gentle ping

Issue #142, opened 2023-06-21
RTL8710BX - authentication error when updating OTA from 2023.5.0-dev to 2023.5.0-dev


mihsu81:
Hi @kuba2k2,

I'm getting the below error when trying to update OTA from libretiny 2023.5.0-dev to libretiny 2023.5.0-dev:
```
INFO Reading configuration /config/libretuya-esphome/plug-ezviz.yaml...
INFO Detected timezone 'Europe/Bucharest'
INFO Generating C++ source...
INFO Compiling app...
Processing plug-ezviz (board: generic-rtl8710bx-4mb-980k; framework: arduino; platform: https://github.com/kuba2k2/libretiny.git)
--------------------------------------------------------------------------------
HARDWARE: RTL8710BX 62MHz, 256KB RAM, 980KB Flash
 - framework-arduino-api @ 2022.8.24+sha.237b10a 
 - framework-realtek-amb1 @ 0.0.0+v2022.06.21.sha.c4e44ef 
 - library-flashdb @ 1.2.0+sha.d5c892f 
 - library-freertos @ 8.1.2+sha.776ae6c 
 - library-freertos-port @ 2023.3.13+sha.bd96e82 
 - library-lwip @ 2.1.3-amb1+sha.6297b80 
 - library-printf @ 6.1.0+sha.28a79bd 
 - tool-ltchiptool @ 4.0.0+sha.788ba4e 
PLATFORM VERSIONS:
 - libretiny @ 1.0.2+sha.5c4da6e
 - ltchiptool @ 4.2.2
CUSTOM OPTIONS:
 - fw_name = esphome
 - fw_version = 2023.5.0-dev
Dependency Graph
|-- ESPAsyncWebServer-esphome @ 3.0.0+sha.9f822c0
|   |-- AsyncTCP-esphome @ 2.0.0+sha.aab1fe4
|-- DNSServer @ 1.1.0
|-- noise-c @ 0.1.4
|   |-- libsodium @ 1.10018.1
|-- ArduinoJson @ 6.18.5
RAM:   [==        ]  20.6% (used 54068 bytes from 262144 bytes)
Flash: [=====     ]  52.3% (used 524804 bytes from 1003520 bytes)
========================= [SUCCESS] Took 9.34 seconds =========================
INFO Successfully compiled program.
INFO Resolving IP address of plug-ezviz.local
INFO  -> 192.168.45.123
[...]

PR #156, opened 2023-08-16
[realtek-ambz] Enable Mbed-TLS for MD5 hashing, remove Polar SSL

hn:
Set Mbed-TLS as default for MD5 hashing, remove Polar SSL. The latter has given wrong results on at least one MCU and is generally deprecated.

Closes issue #142.


kuba2k2:
Well, I think I found the issue. The PolarSSL MD5 context in LT is defined as follows:
```cpp
typedef struct {
	unsigned long total[2];	  /*!< number of bytes processed  */
	unsigned long state[4];	  /*!< intermediate digest state  */
	unsigned char buffer[64]; /*!< data block being processed */
} md5_context;
```
whereas the context in PolarSSL library is this:
```cpp
typedef struct
{
    uint32_t total[2];          /*!< number of bytes processed  */
    uint32_t state[4];          /*!< intermediate digest state  */
    unsigned char buffer[64];   /*!< data block being processed */

    unsigned char ipad[64];     /*!< HMAC: inner padding        */
    unsigned char opad[64];     /*!< HMAC: outer padding        */
}
md5_context;
```

Notice the missing `ipad` and `opad` fields, which make the structure smaller (thus, esphome allocates less memory for the context, and we have a memory corruption).

Issue #153, opened 2023-08-04
BL0942 Reading from UART timed out 


theanch92:
Hello, with version 1.2.0 of libretiny  there is a bug that does not allow the correct reading of data from UART BL0942, while with version 1.1.0 the data are read correctly.

The device is Aubess power monitor switch.

This is the log on **1.2.0**:

> 23:18:50    [E]    [uart:015] Reading from UART timed out at byte 1!
23:18:50    [W]    [bl0942:043] Junk on wire. Throwing away partial message
23:18:50    [D]    [uart_debug:114] >>> 58:AA
23:18:50    [E]    [uart:015] Reading from UART timed out at byte 0!
23:18:50    [D]    [uart_debug:114] <<< 55:00:00:00:4A:5A:38:50:B9:00:00:00:00:00:00:00:26:4E:00:22:01:00:D6

This is the log on **1.1.0**:

> 09:44:43	[D]	[uart_debug:114] >>> 58:AA
> 09:44:43	[D]	[sensor:093] 'BL0942 Voltage': Sending state 232.27730 V with 1 decimals of accuracy
> 09:44:43	[D]	[sensor:093] 'BL0942 Current': Sending state 0.00013 A with 2 decimals of accuracy
> 09:44:43	[D]	[sensor:093] 'BL0942 Power': Sending state 0.00671 W with 0 decimals of accuracy
> 09:44:43	[D]	[sensor:093] 'BL0942 Energy': Sending state 0.00000 kWh with 0 decimals of accuracy
> 09:44:43	[D]	[sensor:093] 'BL0942 Frequency': Sending state 50.00500 Hz with 2 decimals of accuracy
> 09:44:43	[D]	[uart_debug:114] <<< 55:21:00:00:6D:42:38:60:B9:00:FC:FF:FF:00:00:00:1E:4E:00:02:01:00:C8
> 

Configuration:

> esphome:
>   name: cb2s
>   name_add_mac_suffix: true
> 
> libretiny:
>   board: generic-bk7231n-qfn32-tuya
>   framework:
>     version: dev
>
> sensor:
> platform: uptime
> name: Uptime
> platform: bl0942
> uart_id: uart_bus
> current:
>   name: BL0942 Current
> voltage:
>   name: BL0942 Voltage
> power:
>   name: BL0942 Power
>   filters:
>     multiply: -1
> energy:
>   name: BL0942 Energy
> frequency:
>   name: BL0942 Frequency
[...]

Issue #155, opened 2023-08-11
HLW8012... W [      0.102] CHANGE interrupts not supported


alexxx113:

kuba2k2:
Please uninstall all versions of LibreTiny from .platformio directory. Your dev version is too old, that's why setting 1.1.0 manually works.

alexxx113:
Can I help?
logs?
tests?

kuba2k2:
Not really. Seems like CHANGE interrupts were "supported" previously, only due to a bug that produced undefined behavior. Realtek doesn't support CHANGE interrupts properly and that's why it's failing now.

When you tried 1.1.0, did it actually report correct energy measurements? Or just produced no warning?

alexxx113:
version: 1.1.0
work no problem

sensor:
  - platform: hlw8012
    model: BL0937


log

01:12:01 | [D] | [sensor:093] | 'Voltage': Sending state 225.81564 V with 1 decimals of accuracy
01:12:01 | [D] | [sensor:093] | 'Power': Sending state 25.43914 W with 1 decimals of accuracy
01:12:01 | [D] | [sensor:093] | 'Energy': Sending state 541.63861 Wh with 1 decimals of accuracy
01:12:06 | [D] | [hlw8012:082] | Got power=22.6W, voltage=226.0V
01:12:06 | [D] | [sensor:093] | 'Voltage': Sending state 226.04918 V with 1 decimals of accuracy
01:12:06 | [D] | [sensor:093] | 'Power': Sending state 22.57724 W with 1 decimals of accuracy
01:12:06 | [D] | [sensor:093] | 'Energy': Sending state 541.66998 Wh with 1 decimals of accuracy

mihsu81:
My guess is that it is related to the change introduced in 1.2.0 which adds preliminary support for RTL8720C.

https://github.com/kuba2k2/libretiny/blob/f3871388ce54bf277f2067370ee99c57b677e186/cores/realtek-amb/arduino/src/wiring_irq.c#L65

SuperXL2023:
I found a temporary solution.
Slightly modified the code in the file: < config dir >/.esphome/platformio/platforms/libretiny/cores/realtek-amb/arduino/src/wiring_irq.c

```
file: wiring_irq.c

...<other code> 
62: #if LT_RTL8720C
63:                        event = IRQ_FALL_RISE; // <- BUG  [ IRQ_FALL_RISE ] not defined, the compiler stops at this point with an error
[...]

PR #173, opened 2023-09-20
[libs] Fix SerialClass available() return value 

descipher:
This PR changes the SerialClass::available() return value to 0 if no data is available on the referenced UART port. 
Issues:
#153
#155


catalin2402 (approved):

kuba2k2:
Closes #153.

PR #220, opened 2023-12-08
[beken-72xx] Improve ddev ADC support

szupi-ipuzs:
One of the users reported issue esphome/issues#5201 with ADC on BK7231N, so I took a look at the code and found something that could be related. I don't own any N device with ADC, so cannot really test this. But I have checked my T devices with this fix and the ADC still works there.

### DIFF
--- a/cores/beken-72xx/arduino/src/wiring_analog.c
+++ b/cores/beken-72xx/arduino/src/wiring_analog.c
@@ -66,13 +66,21 @@ uint16_t analogReadVoltage(pin_size_t pinNumber) {
 	adc.pData		   = adcData;
 	adc.data_buff_size = 1;
 	handle			   = ddev_open(SARADC_DEV_NAME, &status, (uint32_t)&adc);
-	if (status)
+	if (handle == -1) {
 		return 0;
+	}
+
+	if (status != SARADC_SUCCESS) {
+		ddev_close(handle);
+		return 0;
+	}
+
 	// wait for data
 	while (!adc.has_data || adc.current_sample_data_cnt < 1) {
 		delay(1);
 	}
-	ddev_control(handle, SARADC_CMD_RUN_OR_STOP_ADC, (void *)false);
+	uint8_t run_stop = 0; // stop
+	ddev_control(handle, SARADC_CMD_RUN_OR_STOP_ADC, &run_stop);
 	ddev_close(handle);
 	return adcData[0];
 }


cores/beken-72xx/arduino/src/wiring_analog.c
```diff
@@ -66,13 +66,21 @@ uint16_t analogReadVoltage(pin_size_t pinNumber) {
 	adc.pData		   = adcData;
 	adc.data_buff_size = 1;
 	handle			   = ddev_open(SARADC_DEV_NAME, &status, (uint32_t)&adc);
-	if (status)
+	if (handle == -1) {
 		return 0;
+	}
+
+	if (status != SARADC_SUCCESS) {
+		ddev_close(handle);
```
szupi-ipuzs, line 74: This seems to be important, as open() enables something (CMD_SCTRL_BLK_ENABLE) and can _later_ return error via status. close() disables this "something", so it must be called even if status is error.

cores/beken-72xx/arduino/src/wiring_analog.c
```diff
@@ -66,13 +66,21 @@ uint16_t analogReadVoltage(pin_size_t pinNumber) {
...
+		return 0;
+	}
+
 	// wait for data
 	while (!adc.has_data || adc.current_sample_data_cnt < 1) {
 		delay(1);
 	}
-	ddev_control(handle, SARADC_CMD_RUN_OR_STOP_ADC, (void *)false);
+	uint8_t run_stop = 0; // stop
```
szupi-ipuzs, line 82: This is also important, as the code handling SARADC_CMD_RUN_OR_STOP_ADC expects a pointer and dereferences it blindly. I'm surprise it did not crash on this (is there another layer that checks pointers validity?). Either way, calling with essentially a nullpointer, might not have had the desired effect (of stopping).

cores/beken-72xx/arduino/src/wiring_analog.c
szupi-ipuzs: Oh, and the status can actually be SARADC_FAILURE if open() was called twice, without close() (based on saradc_is_busy)

cores/beken-72xx/arduino/src/wiring_analog.c
kuba2k2: Oh, nice catch. I don't know how the heck this landed here... this was 1.5 years ago.

On Beken (and other microcontrollers) dereferencing the NULL pointer is perfectly valid - it just reads memory at 0x0. On BK7231 it happens to be the bootloader, on Realtek it's the ROM, and so on. Since *0x0 is never 0, the code must have never stopped the ADC...

szupi-ipuzs:
> I wish we could just `#define static ;` or something haha

That would be really confusing :)
But I don't understand the problem - why can't we ditch these functions and create a nice wrapper that would operate on registers directly? Is it simply too much work or is there something missing?

kuba2k2:
Of course we can. This is just too much work for a single person :)

Ultimately (in distant future) I would like to have LibreTiny C APIs (not Arduino APIs) around all peripherals that are now Arduino-only, such as GPIO, ADC, PWM, I2C, SPI, Wi-Fi.

The existing Arduino implementations would then be moved to "components" (similar to those in esp-idf, structured like the `lt_*.c` files now) in the C API. The Arduino APIs would only be a thin wrapper around the C API.

Those "components" would then use plain register access, slowly getting rid of parts of the vendor SDKs. But those are dreams only :smile:

szupi-ipuzs:
Too much work for a single person who does it in their free time and not for money. That's true.
But it's good you have dreams, because it means the software you create is (or will be) well-thought and structured, unlike some others I have seen.
Let's just hope tuya doesn't soon switch their products to yet another chip family - with yet uglier sdks :)

kuba2k2:
> Too much work for a single person who does it in their free time and not for money. 

Yup, that's me :smile:

We still have a few more chip families to add...

kuba2k2 (approved):

PR #222, opened 2023-12-12
[beken-72xx] Pause PWM instead of stopping, track PWM state

szupi-ipuzs:
This seems to resolve issues that occured when pwm was stopped on N-chips.
Instead of stopping, we should do pause (ie. "unit disable"), when duty cycle is 0.
I couldn't make it work using "CMD_PWM_UNIT_DISABLE" and then "CMD_PWM_UNIT_ENABLE", as "disabling" seems to do some additional cleanup for some reason. Although it seems to be working fine with "CMD_PWM_INIT_PARAM" instead of "enabling".

I decided it will be easier to track the state as an enum, so that now it's clear when we should (re)start and/or pause or do full stop.


szupi-ipuzs:
I was finally able to test this PR with a T-device and PWM seems to work fine there too.
Do you need need anything else to merge this?

kuba2k2 (approved):

PR #226, opened 2023-12-23
[beken-72xx] Free list returned by wlan_sta_scan_result()

szupi-ipuzs:
This method might return a dynamically allocated buffer which should be freed by the caller (the example code does it). Not doing this results in memory leaks.

### DIFF
--- a/cores/beken-72xx/arduino/libraries/WiFi/WiFiScan.cpp
+++ b/cores/beken-72xx/arduino/libraries/WiFi/WiFiScan.cpp
@@ -16,6 +16,8 @@ static void scanHandler(void *ctx, uint8_t param) {
 	}
 
 	ScanResult_adv result;
+	result.ApNum  = 0;
+	result.ApList = NULL;
 	if (wlan_sta_scan_result(&result)) {
 		LT_EM(WIFI, "Failed to get scan result");
 		goto end;
@@ -46,6 +48,9 @@ end:
 		// running == false means it was discarded (timeout)
 		scan->running = false;
 		xSemaphoreGive(cDATA->scanSem);
+	}
+	if (result.ApList) {
+		free(result.ApList);
 	}
 	LT_HEAP_I();
 	return;
--- a/cores/beken-72xx/arduino/libraries/WiFi/WiFiScan.cpp
+++ b/cores/beken-72xx/arduino/libraries/WiFi/WiFiScan.cpp
@@ -24,12 +24,13 @@ static void scanHandler(void *ctx, uint8_t param) {
 	}
 	LT_IM(WIFI, "Found %d APs", result.ApNum);
 
-	if (!cls->scanAlloc(result.ApNum)) {
+	auto apNum = cls->scanAlloc(result.ApNum);
+	if (0 == apNum) {
 		LT_WM(WIFI, "scan->ap alloc failed");
 		goto end;
 	}
 
-	for (uint8_t i = 0; i < result.ApNum; i++) {
+	for (uint8_t i = 0; i < apNum; i++) {
 		scan->ap[i].ssid	= strdup(result.ApList[i].ssid);
 		scan->ap[i].auth	= securityTypeToAuthMode(result.ApList[i].security);
 		scan->ap[i].rssi	= result.ApList[i].ApPower;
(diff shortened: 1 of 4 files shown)


cores/beken-72xx/arduino/libraries/WiFi/WiFiScan.cpp
```diff
@@ -47,6 +49,9 @@ static void scanHandler(void *ctx, uint8_t param) {
 		scan->running = false;
 		xSemaphoreGive(cDATA->scanSem);
 	}
+	if (result.ApList) {
+		free(result.ApList);
```
szupi-ipuzs, line 53: The sdk code allocates this mem using some os-independent wrappers. I'm pretty sure they all and up calling malloc() or calloc(), so its safe to call free() on it - but please someone double check it.

cores/realtek-amb/arduino/libraries/WiFi/WiFiScan.cpp
```diff
@@ -20,7 +20,10 @@ static rtw_result_t scanHandler(rtw_scan_handler_result_t *result) {
 	if (!net->SSID.len)
 		return RTW_SUCCESS;
```
szupi-ipuzs, line 21: I can see this function always returns RTW_SUCCESS, even on error. Is this intended?

cores/beken-72xx/arduino/libraries/WiFi/WiFiScan.cpp
kuba2k2: Yes, all wrappers I've seen so far use FreeRTOS eventually. And the standard library memmang functions are linked to FreeRTOS as well, so it's perfectly safe to use `free()` here.

cores/common/arduino/libraries/api/WiFi/WiFi.h
```diff
@@ -163,7 +163,7 @@ class WiFiClass {
 	);
 
 	int16_t scanComplete();
-	uint8_t scanAlloc(uint8_t count);
```
kuba2k2, line 166: These are Arduino APIs, we're not supposed to change the function signatures or behavior. As unfortunate as it is (the APIs are pretty badly designed IMHO), doing otherwise may cause undesired problems with compilation if the user's code uses these methods for some reason.

cores/realtek-amb/arduino/libraries/WiFi/WiFiScan.cpp
kuba2k2: Probably; this is a scanning callback called by the SDK, I don't even think it cares about errors. So the return value is most likely meaningless, and returning an error code would likely cause some other problems.

kuba2k2 (changes requested):

cores/common/arduino/libraries/api/WiFi/WiFi.h
szupi-ipuzs: I was afraid of something like that.
I do not see any description of the return value, but returning the previous count does not allow this method to indicate an error. And, correct me if i'm wrong, failure to allocate a memory should be quite common on systems like that, so it is important to know that and react (eg. by retrying or at least not writing to the memory area).
So, I would strongly suggest changing the API even risking breaking the user's code.

Alternatively, if it is only the method's signature that must remain the same, I suggest returning instead the count of items that we successfully allocated. Then the user can compare it to what was requested and judge by that. What do you think?

cores/realtek-amb/arduino/libraries/WiFi/WiFiScan.cpp
szupi-ipuzs: I've just checked and it appears that you're right - the SDK completely ignores the value returned by this callback. So for the sake of consistency it's better to always return RTW_SUCCESS.

cores/common/arduino/libraries/api/WiFi/WiFi.h
kuba2k2: Actually, upon checking I see that there's no such method in ESP32 at all:
https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/src/WiFiScan.h

Sorry for the confusion. I must have taken the wifi library from somewhere else then, because I don't think I'd ever write a function with such a weird return value.

Of course, you're free to keep your changes in this case. This function could even be made private now - no other class is supposed to use it anyway.

szupi-ipuzs:
Ok, as you can see I decided to make the scanAlloc() method return the number of succesfully allocated items. This can help if realloc() fails - then we still can use the previously allocated buffer if needed.
Making this method private should be done as a separate change, since it probably isn't the only one that should be moved.

I have tested this code shortly on bk7231, but not on realtek, since I do not own any realtek device atm (only checked compilation).

kuba2k2 (approved):

PR #228, opened 2023-12-25
Update esphome.md

daniel-dona:
With the "docker pull" part docker-compose gives this error:

```
ERROR: invalid reference format
```


daniel-dona:
Oops, just realized is exactly the same as https://github.com/libretiny-eu/libretiny/pull/225

Issue #230, opened 2023-12-29
Deep sleep using timer limited to 2 minutes


daniel-dona:
Looks like setting the wake up timer for deep sleep only works as expected up to about 120 seconds (~2 min). Setting a longer sleep time results on sleeping  only for about 15 ~ 20 seconds.

With 150 seconds

```
[1703875377] (2023-12-29 19:42:57):  Type [C-a] [C-h] to see available commands
[1703875377] (2023-12-29 19:42:57):  Terminal ready
[1703875381] (2023-12-29 19:43:01):  Boot
[1703875381] (2023-12-29 19:43:01):  Deep sleep in 5
[1703875382] (2023-12-29 19:43:02):  Deep sleep in 4
[1703875383] (2023-12-29 19:43:03):  Deep sleep in 3
[1703875384] (2023-12-29 19:43:04):  Deep sleep in 2
[1703875385] (2023-12-29 19:43:05):  Deep sleep in 1
[1703875386] (2023-12-29 19:43:06):  Deep sleep for: 150 seconds
[1703875405] (2023-12-29 19:43:25):  Boot
[1703875405] (2023-12-29 19:43:25):  Deep sleep in 5
[1703875406] (2023-12-29 19:43:26):  Deep sleep in 4
[1703875407] (2023-12-29 19:43:27):  Deep sleep in 3
[1703875408] (2023-12-29 19:43:28):  Deep sleep in 2
[1703875409] (2023-12-29 19:43:29):  Deep sleep in 1
[1703875410] (2023-12-29 19:43:30):  Deep sleep for: 150 seconds
[1703875430] (2023-12-29 19:43:50):  Boot
[1703875430] (2023-12-29 19:43:50):  Deep sleep in 5
[1703875431] (2023-12-29 19:43:51):  Deep sleep in 4
[1703875432] (2023-12-29 19:43:52):  Deep sleep in 3
[1703875433] (2023-12-29 19:43:53):  Deep sleep in 2
[1703875434] (2023-12-29 19:43:54):  Deep sleep in 1
[1703875435] (2023-12-29 19:43:55):  Deep sleep for: 150 seconds
[1703875454] (2023-12-29 19:44:14):  Boot
```

With 120 seconds

```
[1703875499] (2023-12-29 19:44:59):  Type [C-a] [C-h] to see available commands
[1703875499] (2023-12-29 19:44:59):  Terminal ready
[1703875502] (2023-12-29 19:45:02):  Boot
[...]

PR #232, opened 2024-01-01
[beken-72xx] Fix deep_sleep time parameter overflow

daniel-dona:
Fix for #230

### DIFF
--- a/cores/beken-72xx/base/api/lt_sleep.c
+++ b/cores/beken-72xx/base/api/lt_sleep.c
@@ -21,12 +21,12 @@ void lt_deep_sleep_unset_gpio(uint32_t gpio_index_map) {
 
 void lt_deep_sleep_config_timer(uint32_t sleep_duration) {
 	deep_sleep_param.wake_up_way |= PS_DEEP_WAKEUP_RTC;
-	uint64_t duration_math = 32768 * sleep_duration;
-	if (duration_math / 1000 > 0xFFFFFFFF) {
+	uint32_t duration_math = (32768 * ((float) sleep_duration)/1000.0);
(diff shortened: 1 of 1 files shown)


cores/beken-72xx/base/api/lt_sleep.c
```diff
@@ -21,12 +21,12 @@ void lt_deep_sleep_unset_gpio(uint32_t gpio_index_map) {
 
 void lt_deep_sleep_config_timer(uint32_t sleep_duration) {
 	deep_sleep_param.wake_up_way |= PS_DEEP_WAKEUP_RTC;
-	uint64_t duration_math = 32768 * sleep_duration;
-	if (duration_math / 1000 > 0xFFFFFFFF) {
+	uint32_t duration_math = (32768 * ((float) sleep_duration)/1000.0);
```
kuba2k2, line 24: To avoid unnecessary casting, could you try `32.768 * sleep_duration`? This way the division is avoided, and casting is implicit because the first operand is a float.

cores/beken-72xx/base/api/lt_sleep.c
```diff
@@ -21,12 +21,12 @@ void lt_deep_sleep_unset_gpio(uint32_t gpio_index_map) {
...
+	if (duration_math > 0xFFFFFFFF) {
```
kuba2k2, line 25: This can never be larger than 0xFFFFFFFF now. Is that even needed?

kuba2k2 (changes requested):

Cossid:
This can probably be closed as duplicate of the now-completed #253

PR #252, opened 2024-02-19
Update Readme & improve md layout

HumbleDeer:
Made some improvements to the markdown text markup, as well as updated some text to reflect the current state of the project.

- ESPHome now supports LibreTiny natively
- Added extra PIO docs references
- Reworded some sections slightly, to accomodate proper grammar
- PIO deprecated the `platformio platform install` command, replaced this with the official recommendation/replacement
- Replaced hard tabs with space characters. These break things less often, and are strongly preferred.

## Tips for others:

- A line break without creating a new paragraph can be accomplished by ending your line with two whitespace characters. You can see this being used in some of the unordered lists.
- Use 4 space characters instead of tabs. Tabs have a tendency to display or behave inconsistently. This difference is already very visible between the GitHub web editor and Visual Studio Code.
- Use the Prettier or Markdown All-In-One plugins for VS Code when writing Markdown. It helps you stay in line with the markdown writing guidelines. 

## Remarks:

The file lacks a trailing space character, which is required by the Markdown spec. This space character seems to get removed whenever something is committed through the GitHub web editor. I can't add it without spinning up a clone of the repo, which is beyond what I have time for at this very moment.

### DIFF
--- a/docs/getting-started/README.md
+++ b/docs/getting-started/README.md
@@ -3,46 +3,59 @@
 Using LibreTiny is simple, just like every other PlatformIO development platform.
 
 1. [Install PlatformIO](https://platformio.org/platformio-ide)
-2. `platformio platform install -f https://github.com/libretiny-eu/libretiny`
+2. `pio pkg install --platform https://github.com/libretiny-eu/libretiny.git`
 
 !!! tip
-	See the [Cloudcutter video guide](https://www.youtube.com/watch?v=sSj8f-HCHQ0) for a complete tutorial on flashing with [Cloudcutter](https://github.com/tuya-cloudcutter/tuya-cloudcutter) and installing [LibreTiny-ESPHome](../projects/esphome.md). **Includes Home Assistant Add-On setup.**
+    See the [Cloudcutter video guide](https://www.youtube.com/watch?v=sSj8f-HCHQ0) for a complete tutorial on flashing with [Cloudcutter](https://github.com/tuya-cloudcutter/tuya-cloudcutter) and installing [LibreTiny-ESPHome](../projects/esphome.md). **Includes Home Assistant Add-On setup.**
+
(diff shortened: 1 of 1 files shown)


HumbleDeer:
You can use shift key and control key as modifiers for the backspace key.
They remove entire lines or entire words/whitespace separated sections.

If I said to leave a space character at the end of the file, I did mean a
newline. The git diff on GitHub didn't show a newline, so I was under the
impression that it had removed it in the process of submitting.

The same reasons you specify for not liking space characters instead of tab
characters is the same set of reasons programmers often prefer space
characters. They are very deterministic and have a predictable effect.

Vs code has a setting bottom right where you can define what you'd like to
use as a tab. You can set it to a tab, or anywhere from one to four space
characters. Vs code will treat tabs created with spaces as tabs and they
will delete / indent as expected.

Cheers. :)

kuba2k2:
> They are very deterministic and have a predictable effect.

Well, the effect will be the same even if we use 5 spaces to indent lines :smile: MkDocs won't care anyway, HTML doesn't render anything of that. So I guess this just affects writing docs, not viewing them.

> Vs code has a setting bottom right where you can define what you'd like to
use as a tab. 

It does, but if a particular project is indented with 2 spaces (ESPHome in this case), I am forced to see it like that, and I can't change the width to 4 [characters] to make the code look less squished. Using tabs I can choose whatever I want on my end, each developer can do what they prefer.

> If I said to leave a space character at the end of the file, I did mean a
newline. The git diff on GitHub didn't show a newline, so I was under the
impression that it had removed it in the process of submitting.

It does indeed keep the newline. In fact, if there's no newline it will show :no_entry: icon at the end.

HumbleDeer:
Pydown definitely does care about indentation! It's an important hierarchy
indicator.

PR #260, opened 2024-03-01
[libs] Fix mDNS string memory corruption, print error on record add failure

szupi-ipuzs:
strcpy() always adds null-terminator at the end.
The original code allocated one byte less than what is needed, so it was overwriting _something_. Could've been important or not :)
Found this while debugging #248, but I don't know if it fixes it or not. In my local tests on ubuntu I was not able to reproduce the behavior described in #248. But if it's memory related, then it's bound to behave differently on x86 + linux. Worth checking if it helped.

### DIFF
--- a/cores/common/arduino/libraries/common/mDNS/mDNS.cpp
+++ b/cores/common/arduino/libraries/common/mDNS/mDNS.cpp
@@ -2,12 +2,21 @@
 
 #include "mDNS.h"
 
-static char *ensureUnderscore(const char *value) {
-	uint8_t len	 = strlen(value) + (value[0] != '_') + 1; // 1 for underscore (if needed), 1 for null-terminator
+static const char *ensureUnderscore(const char *value) {
+	if (value[0] == '_') {
+		return value;
+	}
+	size_t len	 = strlen(value) + 1 + 1; // 1 for underscore, 1 for null-terminator
 	char *result = (char *)malloc(len);
 	result[0]	 = '_';
-	strcpy(result + 1, value + (value[0] == '_'));
+	strcpy(result + 1, value + 1);
 	return result;
+}
+
+static void freeIfDuplicate(const char *original, const char *duplicate) {
+	if ((duplicate) && (original != duplicate)) {
+		free((void *)duplicate);
+	}
 }
 
 void mDNS::setInstanceName(const char *name) {
@@ -21,7 +30,7 @@ bool mDNS::addService(char *service, char *proto, uint16_t port) {
 	uint8_t _proto = strncmp(proto + (proto[0] == '_'), "tcp", 3) == 0 ? MDNS_TCP : MDNS_UDP;
 
 	bool result = addServiceImpl(instanceName ? instanceName : "LT mDNS", _service, _proto, port);
-	free(_service);
+	freeIfDuplicate(service, _service);
 	return result;
 }
 
@@ -34,7 +43,7 @@ bool mDNS::addServiceTxt(char *service, char *proto, char *key, char *value) {
 	sprintf(txt, "%s=%s", key, value);
 
 	bool result = addServiceTxtImpl(_service, _proto, txt);
-	free(_service);
+	freeIfDuplicate(service, _service);
 	free(txt);
 	return result;
 }
--- a/cores/common/arduino/libraries/common/mDNS/mDNS.cpp
+++ b/cores/common/arduino/libraries/common/mDNS/mDNS.cpp
@@ -15,7 +15,7 @@ static char *ensureUnderscore(char *value) {
 
 static inline void freeIfCopied(const char *original, char *duplicate) {
 	if ((duplicate) && (original != duplicate)) {
-		free((void *)duplicate);
+		free(duplicate);
 	}
 }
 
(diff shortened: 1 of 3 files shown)


cores/common/arduino/libraries/common/mDNS/mDNS.cpp
```diff
@@ -3,7 +3,7 @@
 #include "mDNS.h"
 
 static char *ensureUnderscore(const char *value) {
-	uint8_t len	 = strlen(value) + 1;
+	uint8_t len	 = strlen(value) + 1 + 1; // 1 for underscore, 1 for null-terminator
```
Cossid, line 6: Should this be conditional on `value[0] == '_'` (or `!=`) like below for that extra 1?
Granted malloc'ing 1 extra byte isnt the end of the world, and it's free'd relatively soon after anyway.

Cossid: yeah, `if (value[0] == '_') return value;` seems like the better optimization.

cores/common/arduino/libraries/common/mDNS/mDNS.cpp
```diff
@@ -2,14 +2,23 @@
...
+	}
+	size_t len	 = strlen(value) + 1 + 1; // 1 for underscore, 1 for null-terminator
 	char *result = (char *)malloc(len);
 	result[0]	 = '_';
-	strcpy(result + 1, value + (value[0] == '_'));
+	strcpy(result + 1, value);
 	return result;
 }
 
+static void freeIfDuplicate(const char *original, const char *duplicate) {
```
kuba2k2, line 16: A better name could be `freeIfCopied` - I only realized why it frees if `original != duplicate` after a few minutes. Somehow "duplicate" makes me think of it as "equal" or "the same".

Also, try making it inline for performance reasons.

kuba2k2 (changes requested):

cores/common/arduino/libraries/common/mDNS/mDNS.cpp
szupi-ipuzs: I misspelled, I should've named it duplicated, which is a synonim for copied. But copied is fine too.

cores/common/arduino/libraries/common/mDNS/mDNS.cpp
```diff
@@ -2,39 +2,48 @@
...
+	}
+	size_t len	 = strlen(value) + 1 + 1; // 1 for underscore, 1 for null-terminator
 	char *result = (char *)malloc(len);
 	result[0]	 = '_';
-	strcpy(result + 1, value + (value[0] == '_'));
+	strcpy(result + 1, value);
 	return result;
 }
 
+static inline void freeIfCopied(const char *original, const char *duplicate) {
```
kuba2k2, line 16: I'm not sure if it makes sense to use `const char*` in this unit. Freeing const pointers is forbidden, hence the need for a type cast. These strings are allocated on the heap, they're not stored in code memory or R/O memory.

kuba2k2 (changes requested):

cores/common/arduino/libraries/common/mDNS/mDNS.cpp
szupi-ipuzs: According to clean code you should really const every parameter that is not supposed to be changed by the method. And you're right, for the "duplicate" the const doesn't make sense, as it actually _is_ modified (ie. freed).
But "original" should not be touched, so I would leave the other const there.

szupi-ipuzs: By the way, since this is c++ why all this playing with manually allocated c-strings? This is very error-prone. Doesn't arduino have a String class?

kuba2k2: It does, but I believe in Holy-C

Jokes aside, Arduino Strings tend to be a huge performance and RAM bottleneck. I'd even be better off using C++ strings than Arduino strings - heck, even the feature set of these isn't enforced and constant among platforms.

At some point, LibreTiny will get a real C++ low-level API, then there'll be a good time to play with modern C++ features :smiley: The Arduino APIs in LibreTiny are unfortunately meant to be "compatible" (with software that depends on it), not "good".

szupi-ipuzs:
Ok, I think this is enough improvements for now. The only thing left would be to call the new cleanup() method somewhere. In current implementation it will only be called in destructor, but since this is a static object that would be... never?
Shouldn't the mDNS.end() be called in MDNSComponent::on_shutdown() (in mdns_libretiny.cpp)? That would call cleanup right about when it should be, right? Currently it isn't called at all if I'm not mistaken.

kuba2k2:
That's the problem in embedded/IoT programming - these devices most often run as long as they're given power. There's no shutdown, no close button. ESPHome probably never actually calls the shutdown method under normal circumstances, because why would it?
The point is, with the current Arduino API, there's only so much you can improve. I'd say it's okay the way it is and that there's no need to worry about this.

szupi-ipuzs:
I work in embedded for some time now and I've seen this "we won't need a shutdown" approach many times. And in many cases we eventually did need it :), only by then it was very hard to achieve due to many decisions like not doing proper deregistration or deallocations, using static objects and singletons.

And the reasons for needing this in embedded is one: to prevent a full reboot, which usually comes with a price of having to re-initialize and re-register everything, while the only thing you wanted was to reinitialize one small component.

The fact that esphome component does have a shutdown method means they are thinking about having this, so I would say it's worth to be prepared :)

kuba2k2:
Right, but that's a topic for another PR anyway :slightly_smiling_face:

kuba2k2 (approved):

PR #263, opened 2024-03-10
Mdns more fixes

szupi-ipuzs:
I have noticed that if a service is not added to any net_if at the initial call to addServiceImpl(), then it is not stored at all. This is not correct, because a net_if might be nonexisting/down at the time, but later might be up and announced via mdns_netif_ext_status_callback. So all the services added via addServiceImpl should be cached always.
While fixing this I have noticed that the many static vectors can really be merged into a vector of CachedService structure - it makes usage much easier and it cleans-up after itself.

Tested this with some extent using a unit test and on a real device.


Cossid:
Doesn't solve the other outstanding issues, but appears to function fine in the known-working cases here.

szupi-ipuzs:
Anything else needed to merge it? I use this code on every device I have since this PR was added and I have not noticed any issues.

kuba2k2 (approved):

PR #274, opened 2024-04-03
Fix for a race condition in WiFi connection loop

devgs:
There seems to be the race between the event `RW_EVT_STA_CONNECTED` and an actual valid **SSID** value returned by **BDK**. If even a small delay is injected immediately after the event reception the valid value becomes available. Without this fix, due to a polling nature of `ESPHome WiFiComponent::check_connecting_finished` function may observe the `WiFiSTAConnectStatus::CONNECTED` status but with an empty **SSID** value, leading to `Incomplete connection.` warning and immediate attempt to start another connection, while the current one was actually established.

### DIFF
--- a/cores/beken-72xx/arduino/libraries/WiFi/WiFiEvents.cpp
+++ b/cores/beken-72xx/arduino/libraries/WiFi/WiFiEvents.cpp
@@ -27,6 +27,42 @@ static void wifiEventTask(void *arg) {
 	}
 }
 
+// There is a race condition, when we have an event about a successful
+// connection but no SSID yet returned by BDK. Even a sinlge millisecond
+// delay should prevent this from happening. It's better to waste a bit
+// of time here than to lose a valid connection down the line.
+static String waitForValidSSID(WiFiClass * pWiFi)
+{
+	String result;
+
+	// Read the initial value that might just be available already.
+	result = pWiFi->SSID();
+
+	if (!result.length()) {
+		std::size_t i = 0;
+
+		for (; i < 10; ++i) {
+			// Dealy and query again.
+			delay(1);
+			result = pWiFi->SSID();
+
+			if (result.length()) {
+				LT_DM(WIFI, "Got valid SSID after %u delays", i + 1);
+				break;
+			}
+
+			// It's a good idea to yield.
+			yield();
+		}
+
+		if (!result.length()) {
+			LT_DM(WIFI, "Could not obtain a valid SSID after %u delays", i);
+		}
+	}
+
+	return result;
+}
+
 void wifiEventSendArduino(EventId event) {
 	event = (EventId)(RW_EVT_ARDUINO | event);
 	wifiStatusCallback((rw_evt_type *)&event);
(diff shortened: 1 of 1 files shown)


devgs:
Sorry, my bad. Wasn't aware about `clang-format`.

cores/beken-72xx/arduino/libraries/WiFi/WiFiEvents.cpp
```diff
@@ -27,6 +27,41 @@ static void wifiEventTask(void *arg) {
...
+static String waitForValidSSID(WiFiClass *pWiFi) {
+	String result;
+
+	// Read the initial value that might just be available already.
+	result = pWiFi->SSID();
+
+	if (!result.length()) {
+		std::size_t i = 0;
+
+		for (; i < 10; ++i) {
```
kuba2k2, line 43:
```suggestion
		for (std::size_t i = 0; i < 10; i++) {
```

cores/beken-72xx/arduino/libraries/WiFi/WiFiEvents.cpp
```diff
@@ -27,6 +27,41 @@ static void wifiEventTask(void *arg) {
...
+			// Dealy and query again.
```
kuba2k2, line 44:
```suggestion
			// Delay and query again.
```

cores/beken-72xx/arduino/libraries/WiFi/WiFiEvents.cpp
```diff
@@ -27,6 +27,41 @@ static void wifiEventTask(void *arg) {
 	}
 }
 
+// There is a race condition, when we have an event about a successful
+// connection but no SSID yet returned by BDK. Even a sinlge millisecond
```
kuba2k2, line 31:
```suggestion
// connection but no SSID yet returned by BDK. Even a single millisecond
```

kuba2k2 (approved):

cores/beken-72xx/arduino/libraries/WiFi/WiFiEvents.cpp
```diff
@@ -27,6 +27,39 @@ static void wifiEventTask(void *arg) {
...
+				LT_DM(WIFI, "Got valid SSID after %u delays", i + 1);
+				break;
+			}
+
+			// It's a good idea to yield.
+			yield();
+		}
+
+		if (!result.length()) {
+			LT_WM(WIFI, "Could not obtain a valid SSID after %u delays", i);
```
szupi-ipuzs, line 56: This doesn't not compile, as the i variable is now only visible in the for loop scope.

kuba2k2: Ahh, you're right.... of course I had to mess it up.

cores/beken-72xx/arduino/libraries/WiFi/WiFiEvents.cpp
```diff
@@ -27,6 +27,39 @@ static void wifiEventTask(void *arg) {
...
+// delay should prevent this from happening. It's better to waste a bit
+// of time here than to lose a valid connection down the line.
+static String waitForValidSSID(WiFiClass *pWiFi) {
+	String result;
+
+	// Read the initial value that might just be available already.
+	result = pWiFi->SSID();
+
+	if (!result.length()) {
+		for (std::size_t i = 0; i < 10; i++) {
```
Cossid, line 41:
```suggestion
		std::size_t i = 0;
		for (; i < 10; i++) {
```

Cossid (approved):

kuba2k2 (approved):

Issue #44, opened 2022-11-23
Support for RTL8710CM?


craggyh:
Is there support for RTL8710CM yet?
I have a couple of devices (Tapo L920 LED controllers) using this chip and would love to be able to flash them with libretuya/esphome.

kuba2k2:
No, not yet. See #37 and #36.

Nold360:
anything we can do to push support forward? idk what's missing/required :(

Rain:
Thanks for the information!

Yes, I realized support for the RTL8720C family is in the very early stages. Just figured I'd report my findings.

That makes sense regarding the AT CLI. Since the stock TP-Link EP25 had it I assumed it was standard and always built w/ the RTL MCUs. Their firmware must be based off the "demo" (which makes sense, I guess).

A quick perusal through their stock firmware revealed that there seems to be a way to trigger the SDK-included OTA function remotely. No signatures are checked either (you can send garbage and it will be written as long as the checksum is correct). It should be possible replace the firmware on these smart plugs wirelessly without opening them up! 😎

I'll poke around some more with download mode. I've tried `ping\n` on UART2 but I wasn't aware of the PA0/PA13 combination. These EP25 plugs are hard enough to get open so anything that makes re-flashing easier will be helpful.

kuba2k2:
Eventually OTA will be supported.

Have you dumped the original firmware prior to flashing? If yes, you can always mess with it by just flashing it back.

asozio:
I've got a couple of tp-link KP115 Kasa Smart Wi-Fi Plug (slim energy monitoring) which appear to contain RTL8710CF modules (12-pin pcb mounted through hole on the main board)

![20231020_191713](https://github.com/libretiny-eu/libretiny/assets/48190027/b5973ccf-5ae0-4af7-9051-90c053de4888)
![20231020_190758](https://github.com/libretiny-eu/libretiny/assets/48190027/0924318d-a2b9-4d88-a86f-6bc5ac00f82f)
![20231020_194117](https://github.com/libretiny-eu/libretiny/assets/48190027/81a9fb09-454a-46f0-9b85-c7b4c54a3485)
![20231020_194914](https://github.com/libretiny-eu/libretiny/assets/48190027/6c2bc6c2-8362-434a-b35c-f29de17ea252)

I'm keen to test these out with libretiny if someone has some pointers for how I get started?

kuba2k2:
[...]

PR #299, opened 2024-11-26
[realtek-ambz2] Define PIN_SERIAL0_RX/TX for wbr3

nmschulte:
This is necessary; `WBR3` is not in the `feature/i2c-spi` branch, and this fixes about `feature/realtek-update` without the concerns of `feature/i2c-spi` -- perhaps I misunderstand the relevant comment, however.

https://github.com/libretiny-eu/libretiny/issues/44#issuecomment-1792656300

https://github.com/baltpeter/libretiny/commit/a92983a608836a380adfeb76eaa968f1dcabf27b

### DIFF
--- a/boards/variants/wbr3.h
+++ b/boards/variants/wbr3.h
@@ -28,8 +28,10 @@
 // ------------
 #define PIN_SERIAL0_RX_0 12u // PIN_A12
 #define PIN_SERIAL0_RX_1 13u // PIN_A13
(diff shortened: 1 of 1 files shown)


nmschulte:
In that case, I'd say merging is unnecessary -- it sounds like I should focus my efforts on ESPHome and working against the changes you mention.  Thanks for helping my understanding here.

PR #302, opened 2024-12-03
OTA implementation & fixes for RTL8720CF

prokoma:
I've cleaned the history and removed the code to pull ltchiptool from git. You need to make a release of ltchiptool first before merging this PR and also update the minimum required version in [builder/main.py](https://github.com/libretiny-eu/libretiny/blob/master/builder/main.py#L21).


fcol95:
Hey!
Quite new with all of this, but have a board running a rtl8720cf chip, planning on doing an esphome config for it with "board: generic-rtl8720cf-2mb-992k", but seems OTA is not yet working?
Curious about the status of this? Are you close of having functional OTA for the rtl8720cf? Any way I can help? (Not a lot of experience with libretiny at all, but some in embedded system firmware development and a bit of free time these days!)

prokoma:
Hi @fcol95,
OTA for rtl8720cf is not implemented yet in the official release. My solution works, but the code is not ideal and there are [some open questions](https://github.com/libretiny-eu/ltchiptool/pull/46). 

I've described steps how to install esphome with working OTA in https://github.com/libretiny-eu/libretiny/issues/44#issuecomment-2514974466 . You can skip step 0, because ltchiptool >= 4.11.5 already contains the mentioned fix.

fcol95:
@prokoma thanks for your quick reply, just had deleted my comment when stumbling on your open PR in ltchiptool repo and understanding a bit better the issue! thanks for the link to your custom solution!
I'll give it a go once I finish working on the component driver I need to do for the chip that interfaces with this rtl8720c in my device!

fcol95:
Just tested the OTA through esphome and your "source: https://github.com/prokoma/libretiny#55aacc8" framework, works quite well, thanks for the effort, would be glad to see that merged in the production release soon!

MottiG:
Hi @prokoma,

Thanks for a great contribution! I wonder if there are any plans to merge this PR soon?

bdraco:
Thanks for being open to this. I spent some time today digging into how the RTL8720CF bootloader actually decides which image to boot, and I think I can address some of the concerns from the ltchiptool discussion.

From what I can tell, the bootloader doesn't actually use serial numbers for boot selection. It just checks if the first byte of each image is valid (not inverted). The serial numbers seem to be informational only. The SDK's own `sys_update_ota_set_boot_fw_idx()` uses the same byte inversion approach that prokoma implemented.

So the mechanism in this PR appears to be correct for the hardware. It's not a workaround, it's how Realtek designed it.

The main limitation is there's no automatic rollback if the new image is broken since the bootloader has no fallback mechanism. ESPHome's safe_mode catches most failures, but very early crashes would need UART recovery. That's a hardware limitation though, not something we can fix in software.

PR #307, opened 2024-12-17
[realtek-ambz2] Fix partition table & bootloader flashing via UART

prokoma:
Do not include partition table & bootloader in device schemes (consistent with other supported families). Fix partition table & bootloader marked as `SINGLE` instead of `DUAL_1/2` in the flasher scheme, thus skipped during flashing using `ltchiptool`.


kuba2k2:
Agreed. The "issue" of changing partition layouts over-the-air is a separate matter, it will need to be sorted out later.

kuba2k2 (approved):

PR #312, opened 2025-01-07
[lightning-ln882h] Add support for Lightning LN882H family

lamauny:
I'm happy to contribute to this excellent project by adding basic support for Lightning Ln882h family.

[Supported features](https://github.com/lamauny/libretiny/blob/feature/ln882x/docs/status/supported.md)

Wifi AP is only working in Open Auth mode, and Mixed mode (AP+STA) is not supported

More supported features will be added in the future, but this support is sufficient to pilot the relay and handle button of my aliexpress [smart light module](https://fr.aliexpress.com/item/1005008002622624.html) (kind-of) based on a LN-02 module.

Links to the PR on other repos adding this feature :
- [x] [ltchiptool](https://github.com/libretiny-eu/ltchiptool/pull/52)
- [x] [library-freertos](https://github.com/libretiny-eu/library-freertos/pull/1)
- [x] [library-freertos-port](https://github.com/libretiny-eu/library-freertos-port/pull/1)
- [x] [lwip](https://github.com/libretiny-eu/lwip/pull/4)
- [x] [framework-lightning-ln882h](https://github.com/lamauny/framework-lightning-ln882h)


divadiow:
have you tested this on LN882x? LN882H is a newer family with a different SDK.

lamauny:
> have you tested this on LN882x? LN882H is a newer family with a different SDK.

no, i've only tested on ln882h, but it seems that ln882h is from the ln822x family but in 'H series' (I think it means Hybrid WiFi+BLE), so it's worth testing.

divadiow:
hmm. I thought that too. I have flashed OpenBeken LN882H builds to LN8825B quite a number of times without success.

Are you building separate LN882X binaries or is the output a single LN882 fw?

lamauny:
Hey, I've had a look to the vector table of the 2 chips and they are different, so these chips are definitely not fully compatible.

boards/_base/lightning-ln882x.json
```diff
@@ -0,0 +1,47 @@
+{
+	"build": {
+		"family": "LN882X",
+		"ldscript": "ln882x_bsp.ld",
+		"bootfile": "boot_ln882x.bin",
+		"f_cpu": "160000000L",
+		"prefix": "arm-none-eabi-"
```
kuba2k2, line 7:
```suggestion
	"build": {
		"f_cpu": "160000000L",
		"prefix": "arm-none-eabi-"
```

README.md
```diff
@@ -15,10 +15,11 @@
 
 ![RTL8710BN](https://img.shields.io/badge/-rtl8710bn-blue)
 ![BK7231](https://img.shields.io/badge/-bk7231-blue)
+![LN882H](https://img.shields.io/badge/-ln882h-blue)
 
 </div>
 
-PlatformIO development platform for BK7231 and RTL8710 IoT chips.
+PlatformIO development platform for LN882H, BK7231 and RTL8710 IoT chips.
```
kuba2k2, line 22:
```suggestion
PlatformIO development platform for BK7231 and RTL8710 and LN882H IoT chips.
```

SUMMARY.md
```diff
@@ -19,6 +19,7 @@
 	* [Realtek AmebaZ](docs/platform/realtek-ambz/README.md)
 		* [Debugging](docs/platform/realtek-ambz/debugging.md)
 		* [Exception decoder](docs/platform/realtek-ambz/exception-decoder.md)
+	* [Lightning LN882x - info](docs/platform/lightning-ln882x/README.md)
```
kuba2k2, line 22:
```suggestion
	* [Lightning LN882x](docs/platform/lightning-ln882x/README.md)
```

boards/_base/ic/ln882hk.json
```diff
@@ -0,0 +1,142 @@
...
+				"I2C": ["0_SCL","0_SDA"]
+			},
+			"23": {
+				"C_NAME": "PA_12",
+				"GPIO": "P12",
+				"I2C": ["0_SCL","0_SDA"]
+			},
+			"24": {
+				"C_NAME": "PB_3",
+				"GPIO": "P19",
```
kuba2k2, line 101: Why is this GPIO suddenly P19? Does the chip use PA/PB naming scheme or simply P0..P25 naming?

Some chips, such as Beken, don't have PA/PB ports. But others, such as Realtek, do have PA/PB ports.

Please use the naming that's appropriate for this chip. Correct me if I'm wrong.

Use `"GPIONUM": 41,` to assign a correct number for pins in the PB bank.

boards/_base/lightning-ln882hki.json
```diff
@@ -0,0 +1,31 @@
...
+  },
+  "flash": {
+    "boot":     "0x000000+0x06000",
+    "part_tab": "0x006000+0x01000",
+    "app":      "0x007000+0x12C000",
+    "ota":      "0x133000+0xAA000",
+    "nvds":     "0x1DD000+0x03000",
+    "kv":       "0x1E0000+0x04000",
+    "kvs":      "0x1E4000+0x08000",
+    "user":     "0x1EC000+0x14000"
```
kuba2k2, line 15: Reindent this file with tabs, and remove the alignment in flash layout (it's not needed, and any formatter will remove it anyway).

boards/ln-02.json
```diff
@@ -0,0 +1,22 @@
...
+		"pcb/ln-02-test"
+],
+	"build": {
+		"mcu": "ln882hk",
+		"variant": "ln-02"
+	},
+	"name": "LN-02 Wi-Fi/Ble Module",
+  "url": "https://docs.libretiny.eu/boards/ln-02/",
+  "vendor": "Unknown",
+	"doc": {
```
kuba2k2, line 16: Reformat this file.

boards/wl2s.json
```diff
@@ -0,0 +1,22 @@
...
+],
+	"build": {
+		"mcu": "ln882hk",
+		"variant": "wl2s"
+	},
+	"name": "WL2S Wi-Fi/Ble Module",
+  "url": "https://docs.libretiny.eu/boards/wl2s/",
+  "vendor": "Unknown",
+	"doc": {
+		"fccid": ""
```
kuba2k2, line 17: Reformat this file.

builder/family/lightning-ln882h.py
```diff
@@ -0,0 +1,240 @@
...
+
+# Main firmware outputs and actions
+image_fw = "${BUILD_DIR}/image_firmware-ota-xz-v1.0.bin"
+env.Replace(
+    # linker command
+    LINK='${LTCHIPTOOL} link2bin ${BOARD_JSON} "" ""',
+    # UF2OTA input list
+    UF2OTA=[
+        # ota binary image for device
+        f"{image_fw}=device:ota",
```
kuba2k2, line 238: What about the flasher? This will make the resulting UF2 file unusable with ltchiptool.

cores/common/base/lt_api.h
```diff
@@ -10,15 +10,15 @@
...
 
-#include "api/lt_cpu.h"
-#include "api/lt_device.h"
-#include "api/lt_flash.h"
-#include "api/lt_init.h"
-#include "api/lt_mem.h"
-#include "api/lt_ota.h"
-#include "api/lt_sleep.h"
-#include "api/lt_utils.h"
-#include "api/lt_wdt.h"
```
kuba2k2, line 21: Why is that necessary? The `api` directory is in the same path as `lt_api.h`, so it should use "" to reference these files.

cores/common/base/lt_types.h
```diff
@@ -17,6 +17,7 @@ typedef enum {
 	F_BK7231N  = 0x7B3EF230, // Beken 7231N
 	F_BK7251   = 0x6A82CC42, // Beken 7251/7252
 	F_BL60X	   = 0xDE1270B7, // Boufallo 602
+	F_LN882H   = 0x12345678, // Lightning LN882H TODO
```
kuba2k2, line 20:
```suggestion
	F_LN882H   = 0xA38090A8, // Lightning LN882H
```

docs/platform/lightning-ln882x/debugging.md
```diff
@@ -0,0 +1,66 @@
...
+openocd_interface = <interface name>
+```
+where `<interface name>` is for example `raspberrypi2-native`, `stlink`, etc.
+
+## Remote debugger
+
+Using a Raspberry Pi is probably the easiest option (and cheapest, as everyone has a spare Pi laying around).
+
+!!! 
+	NOT TESTED
```
kuba2k2, line 26: What is not tested? Raspberry Pi instructions are not here, perhaps it's better to remove this section altogether?

docs/platform/lightning-ln882x/debug_openocd.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+#sudo openocd -f /usr/share/openocd/scripts/interface/stlink.cfg -f ../../../cores/lightning-ln882x/misc/ln882x.cfg -c 'reset_config srst_only' $*
+sudo openocd -f /usr/share/openocd/scripts/interface/jlink.cfg -f ../../../cores/lightning-ln882x/misc/ln882x.cfg -c 'reset_config srst_only' $*
```
kuba2k2, line 4: Is this file needed? I believe everyone who used OpenOCD will be able to load a config file.

If this is something specific to LN882H, it should be mentioned in the docs instead of in a separate file (which wouldn't be visible on the docs page).

docs/platform/lightning-ln882x/README.md
```diff
@@ -0,0 +1,5 @@
+# Lightning LN882x
+
+- [LN882H](http://www.lightningsemi.com/ln882h)  : MCU Combo Wi-Fi/BLE IoT
+- [LN8829L](http://www.lightningsemi.com/LN8829L1) : Wi-Fi SOC
+
```
kuba2k2, line 5: This page serves little purpose - it would be useful to put datasheet links, specifications, list known chips and their differences, etc.

See the [beken-72xx](https://docs.libretiny.eu/docs/platform/beken-72xx/) page for an example.

platform.json
```diff
@@ -72,6 +72,22 @@
         }
       }
     },
+    "framework-lightning-ln882h": {
+      "type": "framework",
+      "optional": true,
+      "version": "https://github.com/lamauny/framework-lightning-ln882h",
```
kuba2k2, line 78:
```suggestion
      "version": "https://github.com/libretiny-eu/framework-lightning-ln882h",
```

platform.json
```diff
@@ -80,16 +96,16 @@
     "library-lwip": {
       "type": "framework",
       "optional": true,
-      "base_url": "https://github.com/libretiny-eu/lwip"
+      "base_url": "https://github.com/lamauny/lwip"
```
kuba2k2, line 99:
```suggestion
      "base_url": "https://github.com/libretiny-eu/lwip"
```

platform.json
```diff
@@ -80,16 +96,16 @@
...
     },
     "library-freertos": {
       "type": "framework",
       "optional": true,
-      "base_url": "https://github.com/libretiny-eu/library-freertos"
+      "base_url": "https://github.com/lamauny/library-freertos"
```
kuba2k2, line 104:
```suggestion
      "base_url": "https://github.com/libretiny-eu/library-freertos"
```

platform.json
```diff
@@ -80,16 +96,16 @@
...
     },
     "library-freertos-port": {
       "type": "framework",
-      "version": "https://github.com/libretiny-eu/library-freertos-port#2023.05.23"
+      "version": "https://github.com/lamauny/library-freertos-port"
```
kuba2k2, line 108:
```suggestion
      "version": "https://github.com/libretiny-eu/library-freertos-port#2025.01.18"
```

cores/lightning-ln882h/base/api/lt_ota.c
```diff
@@ -0,0 +1,41 @@
...
+#include <sdk_private.h>
+
+typedef enum {
+	UPG_STATE_DOWNLOAD_ING	= 0u,		  /*< ota image downloading */
+	UPG_STATE_DOWNLOAD_OK	= 1u,		  /*< ota image download finish and check success */
+	UPG_STATE_RESTORE_ING	= 2u,		  /*< ota image restoring */
+	UPG_STATE_RESTORE_OK	= 3u,		  /*< ota image restore finish and check success */
+	UPG_STATE_RESTORE_FILED = 4u,		  /*< ota image restore filed */
+	UPG_STATE_REPORT_OK		= 5u,		  /*< ota remote server was notified of the result successfully */
+	UPG_STATE_PLACE			= 0x12345678, /*< make different IDEs compatible */
```
kuba2k2, line 13: I don't think this is needed here.

cores/lightning-ln882h/base/api/lt_ota.c
```diff
@@ -0,0 +1,41 @@
...
+}
+
+void lt_ota_set_write_protect(uf2_ota_t *uf2) {
+	LT_DM(OTA, "lt_ota_set_write_protect");
+	ln_nvds_set_ota_upg_state(UPG_STATE_DOWNLOAD_ING);
+}
+
+bool lt_ota_switch(bool revert) {
+	LT_DM(OTA, "lt_ota_switch(%d)", revert);
+	ln_nvds_set_ota_upg_state(UPG_STATE_DOWNLOAD_OK);
```
kuba2k2, line 39: Refer to https://github.com/libretiny-eu/libretiny/blob/master/cores/beken-72xx/base/api/lt_ota.c#L29 to add a proper `lt_ota_is_valid()` implementation and use it in `lt_ota_switch()`.

cores/lightning-ln882h/base/fixups/reg_cache.h
```diff
@@ -0,0 +1,3 @@
+#ifndef __cplusplus
+#include_next "reg/reg_cache.h"
+#endif
```
kuba2k2, line 3: What are the `reg_*.h` files for? It seems like they are just forwarding `#include <reg_cache.h>` to `#include <reg/reg_cache.h>`.

If that's the case, why can't the "reg/" directory be added to the global include path?

cores/lightning-ln882h/base/lt_family.h
```diff
@@ -0,0 +1,22 @@
...
+#elif LT_HW_UART2
+#define LT_UART_DEFAULT_PORT 2
+#else
+#error "No serial port is available"
+#endif
+#endif
+
+// Auto-download-reboot detection pattern
+// Link check command (CMD_LinkCheck=0)
+#define LT_UART_ADR_PATTERN 0x01, 0xE0, 0xFC, 0x01, 0x00
```
kuba2k2, line 22: That's a Beken command. If auto-download-reboot is not implemented, simply remove this line and it will stay disabled.

cores/lightning-ln882h/base/fixups/os_queue.h
```diff
@@ -0,0 +1,7 @@
+#ifndef OS_QUEUE_H
```
kuba2k2, line 1: If possible, use `#pragma once` instead.

cores/lightning-ln882h/base/api/lt_init.h
```diff
@@ -0,0 +1,25 @@
...
+ * and shouldn't be called manually.
+ */
+void lt_init_variant();
+
+/**
+ * @brief Initialize the family's Arduino core (optional).
+ * This method is family-specific; the family core can do whatever it wants to.
+ * This method is empty if not implemented, and shouldn't be called manually.
+ */
+void lt_init_arduino();
```
kuba2k2, line 25: This file is in LibreTiny common core. It should not be redefined in family cores.

cores/lightning-ln882h/base/api/lt_init.c
```diff
@@ -0,0 +1,61 @@
...
+	ln_runtime_measure_init();
+
+	// 2. register os heap mem
+	OS_DefineHeapRegions();
+
+	// 3. log init
+	lt_init_log();
+
+	cm_backtrace_init("LibreTiny - LN882H", "HW_V1.0", "SW_V1.0");
+	LT_I("------  LibreTiny - LN882H  ------");
```
kuba2k2, line 33: LibreTiny already has a banner that prints some useful info, including the board name.

cores/lightning-ln882h/arduino/src/wiring_digital.c
```diff
@@ -0,0 +1,67 @@
...
+			return;
+	}
+
+	hal_gpio_init(data->gpio_base, gpio);
+}
+
+void digitalWrite(pin_size_t pinNumber, PinStatus status) {
+	pinCheckGetData(pinNumber, PIN_GPIO, );
+	pinSetOutputPull(pin, data, pinNumber, status);
+	if (!!status)
```
kuba2k2, line 57:
```suggestion
	if (status)
```

cores/lightning-ln882h/arduino/src/wiring_analog.c
```diff
@@ -0,0 +1,81 @@
...
+
+	hal_adc_en(ADC_BASE, HAL_ENABLE);
+
+	hal_adc_start_conv(ADC_BASE);
+
+	while (hal_adc_get_conv_status(ADC_BASE, ch) == 0)
+		;
+
+	ret = hal_adc_get_data(ADC_BASE, ch);
+	LT_D("ADC CH(%d) = %d", ch, ret);
```
kuba2k2, line 72: Debug message unnecessary.

cores/lightning-ln882h/arduino/src/wiring.c
```diff
@@ -0,0 +1,73 @@
...
+		hal_gpio_pin_it_en(data->gpio_base, data->gpio->pin, HAL_DISABLE);
+		pinDisable(pin, PIN_IRQ);
+	}
+	/*	TODO EL
+		if ((mask & PIN_PWM) && (pin->enabled & PIN_PWM)) {
+			pwmout_free(data->pwm);
+			free(data->pwm);
+			pinDisable(pin, PIN_PWM);
+		}
+	*/
```
kuba2k2, line 72: Is this needed? If PWM disabling is not implemented, perhaps a more informative comment would be good. Otherwise someone will be wondering what `TODO EL` means :smile:

cores/lightning-ln882h/arduino/src/lt_defs.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#error "Don't include this file directly"
+
+#define LT_ARD_HAS_WIFI	   1
+#define LT_ARD_HAS_SERIAL  1
+#define LT_ARD_HAS_WIRE	   1
+#define LT_ARD_MD5_MBEDTLS 1
+#define LT_ARD_MD5_HOSTAPD 0
```
kuba2k2, line 9:
```suggestion
#define LT_ARD_MD5_MBEDTLS 1
```

cores/lightning-ln882h/arduino/src/Arduino.h
```diff
@@ -0,0 +1,7 @@
+/* Copyright (c) Etienn Le Cousin 2024-02-11. */
+
+#pragma once
+
+#include_next <Arduino.h>
+
+#include <sdk_private.h>
```
kuba2k2, line 7: This file shouldn't be needed. Also, `sdk_private.h` is not supposed to be included by non-LibreTiny code (hence the name "private").

kuba2k2 (changes requested):

builder/family/lightning-ln882h.py
lamauny: In fact ltchiptool generates several files from raw firmware (raw_firmware.elf):
- raw_firmware.bin
- image_firmware.bin (full raw binary with all partitions)
- image_firmware-ota-xz-v1.0.bin (compressed binary image of app parition, to be flashed in ota partition)
- firmware.bin and firmware.uf2 (UF2 encoded binary of image_firmware-ota-xz-v1.0.bin) This is this file which is used for ota update with esphome or via web page OTA download.

Do I miss something ?

cores/common/base/lt_api.h
lamauny: Please see lt_init.h comment.
For my lt_init.h file to be included in place of the original lt_init header I had to change this, otherwise it is impossible to redeclare the functions.

cores/lightning-ln882h/base/api/lt_init.h
lamauny: In fact I encountered an issue with these functions declared as weak : the compiler never linked my functions.
It seems to be an issue of declaring weak prototypes with static libraries.
I had to change the prototype declaration without the weak attribute to get my functions linked correctly...

cores/lightning-ln882h/base/fixups/reg_cache.h
lamauny: The reg_*.h files from sdk fail to compile with the c++ compiler, then I had to find a tweak to get these files not beeing compiled in c++...

cores/lightning-ln882h/base/api/lt_ota.c
lamauny: Not sure of what you mean.
These flags are used by the boot to know that there is a new downloaded ota firmware which need to be decompressed and flashed in the main app partition.

cores/lightning-ln882h/base/api/lt_ota.c
lamauny: The LN882H has one app partition and one OTA partition :
The ota update process is done by flashing a xz compressed app firmware in the OTA partition, then putting a UPG_STATE_DOWNLOAD_OK flag to nv memory and then reboot.
The boot will see this flag and will decompress the ota firmware and flash it to the main app partition, then boot on it.
Then I don't know what I'm supposed to do in `lt_ota_switch()` ? a reset ?
For `lt_ota_is_valid()` , I don't really know how to check that the ota is correctly flashed and valid.

boards/_base/ic/ln882hk.json
lamauny: You're right, in fact I didn't get the naming PA/PB to work but it was at the beginning of my development, now that I know better the library I'll give a second try.

lamauny (commented):
Many files updated.
Please see my comments on remaining ones.

docs/platform/lightning-ln882x/debug_openocd.sh
lamauny: file removed

docs/platform/lightning-ln882x/README.md
lamauny: file updated

cores/lightning-ln882h/base/lt_family.h
lamauny: ok, removed

cores/lightning-ln882h/base/fixups/os_queue.h
lamauny: done

cores/lightning-ln882h/arduino/src/wiring_analog.c
lamauny: message removed

cores/lightning-ln882h/arduino/src/wiring.c
lamauny: ok, removed

cores/lightning-ln882h/base/api/lt_init.c
lamauny: ok, removed

NonPIayerCharacter:
I'm not a LT dev, but i'm a little familiar with LN882H.

Regarding ota verification, check [this function](https://github.com/lamauny/framework-lightning-ln882h/blob/a4d07261cffb1fea58eb53cda91c63508deeca9f/components/ln_at_cmd/ota/ln_at_cmd_ota.c#L217)

Plus i did a partial PWM implementation for OpenBeken, you can look it up [here](https://github.com/openshwprojects/OpenBK7231T_App/blob/main/src/hal/ln882h/hal_pins_ln882h.c). Partial is because i had some trouble with timers, so only half the channels are available (which is still enough for a rgbcw bulb, so i didn't bother fully fixing it).

Additionally i recommend implementing power saving. The chip is very hot, at least compared to beken. More power-hungry too. And it is usually mandatory on cheap chinese devices to prolong their life. I already had a capacitor die in a smart plug after only a year, where i replaced stock module with an ESP-02S, which is more power-hungry than an RTL module it came with. And there are several mentions of the same problem on elektroda forum, but without replacing the module and running without power save.

lamauny:
> I'm not a LT dev, but i'm a little familiar with LN882H.
> 
> Regarding ota verification, check [this function](https://github.com/lamauny/framework-lightning-ln882h/blob/a4d07261cffb1fea58eb53cda91c63508deeca9f/components/ln_at_cmd/ota/ln_at_cmd_ota.c#L217)
> 
It looks good :-)

> Plus i did a partial PWM implementation for OpenBeken, you can look it up [here](https://github.com/openshwprojects/OpenBK7231T_App/blob/main/src/hal/ln882h/hal_pins_ln882h.c). Partial is because i had some trouble with timers, so only half the channels are available (which is still enough for a rgbcw bulb, so i didn't bother fully fixing it).
> 
Ok I'll have a look!

> Additionally i recommend implementing power saving. The chip is very hot, at least compared to beken. More power-hungry too. And it is usually mandatory on cheap chinese devices to prolong their life. I already had a capacitor die in a smart plug after only a year, where i replaced stock module with an ESP-02S, which is more power-hungry than an RTL module it came with. And there are several mentions of the same problem on elektroda forum, but without replacing the module and running without power save.
>
My chip is not really hot, connected to WiFi I measure 100mA DC on 5V power (0.5W), of course without Relay activated : how much power-hungry is it based on your experience ?

Thanx!

NonPIayerCharacter:
Regarding power consumption, i didn't verify it myself, but saw on the [forum](https://www.elektroda.com/rtvforum/topic4076474.html), basic power save reduces consumption by ~0.2W.
A have only a single LN-02 module, not in any device but soldered to d1 mini with wires and acts as a dev board.
Without powersave it is too hot for me to touch, even from the back. With powersave it is still hot, but i can at least hold a finger there without getting burned.

wiltet:
> Regarding power consumption, i didn't verify it myself, but saw on the [forum](https://www.elektroda.com/rtvforum/topic4076474.html), basic power save reduces consumption by ~0.2W. A have only a single LN-02 module, not in any device but soldered to d1 mini with wires and acts as a dev board. Without powersave it is too hot for me to touch, even from the back. With powersave it is still hot, but i can at least hold a finger there without getting burned.

I can confirm that.
And `powersave` would be very useful if possible to implement.

lamauny:
> > Regarding power consumption, i didn't verify it myself, but saw on the [forum](https://www.elektroda.com/rtvforum/topic4076474.html), basic power save reduces consumption by ~0.2W. A have only a single LN-02 module, not in any device but soldered to d1 mini with wires and acts as a dev board. Without powersave it is too hot for me to touch, even from the back. With powersave it is still hot, but i can at least hold a finger there without getting burned.
> 
> I can confirm that. And `powersave` would be very useful if possible to implement.
> 
> Just mesured a LN-02 (sorry) with OpenBeken: +/-100mA `powersave 0` and +/-60mA `powersave 1`

Hi, powersave is already implemented like OpenBeken's `powersave 1` mode.
With my esphome app, I measure 100mA with wifi:power_save_mode and 110mA without.

wiltet:
> With my esphome app, I measure 100mA with `wifi:power_save_mode` and 110mA without.

This looks more like ` wifi:` `powersave mode:` `light`. Is `high` working as well?

boards/variants/ln-02.c
```diff
@@ -0,0 +1,42 @@
...
+#ifdef LT_VARIANT_INCLUDE
+#include LT_VARIANT_INCLUDE
+#endif
+
+// clang-format off
+PinInfo lt_arduino_pin_info_list[PINS_COUNT] = {
+	// D0: PA11, SD_D1, I2C0_SCL, I2C0_SDA
+	{PA_11, PIN_I2C | PIN_GPIO,                               PIN_NONE, 0},
+	// D1: PB03, ADC5, I2C0_SCL, I2C0_SDA
+	{PB_3,  PIN_I2C | PIN_GPIO,                               PIN_NONE, 0},
```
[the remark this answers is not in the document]
kuba2k2: That's because you haven't assigned an `A#` pin in `ln-02.json`. The `ARD` pin mappings should be assigned sequentially. If a pin supports GPIO, it should have a `D#` mapping. If it also supports ADC, it should also have an `A#` mapping.

This is subject to removal in future versions of LibreTiny, but for now that's all we got.

builder/family/lightning-ln882h.py
kuba2k2: The UF2 file contains binaries for both the flasher and the device. It should both have the raw binary for serial flasher, as well as the OTA image.

The UF2 file is parsed by ltchiptool or the on-device OTA code, and the appropriate image is flashed.

The current solution will only store the OTA image, which will make ltchiptool unable to flash anything via serial.

cores/lightning-ln882h/base/api/lt_ota.c
kuba2k2: I mean that the `UPG_STATE_PLACE` is a dummy flag, it's not necessary for anything.

cores/lightning-ln882h/base/api/lt_ota.c
kuba2k2: It is enough to check the OTA header. We can assume that the presence of the header means it's a valid image.

cores/lightning-ln882h/base/fixups/reg_cache.h
kuba2k2: Perhaps you mean to add `extern "C" {}` block to the lines that include these reg files? That's usually what helps if compilation with C++ fails.

cores/lightning-ln882h/base/api/lt_init.h
kuba2k2: There are no issues with weak functions on other families. Common library prototypes should not be redefined.

cores/common/base/lt_api.h
kuba2k2: Functions should not be redeclared, that's why they are marked as weak. I see no reason why would the compiler just ignore the implemented functions, if there are no other implementations.

kuba2k2 (changes requested):

boards/_base/ic/ln882hk.json
```diff
@@ -0,0 +1,149 @@
...
+			"7": {
+				"IO": "I",
+				"CTRL": "CEN"
+			},
+			"8": {
+				"C_NAME": "PA_0",
+				"GPIO": "PA00",
+				"IRQ": null,
+				"ADC": 2,
+				"I2C": ["0_SCL","0_SDA"]
```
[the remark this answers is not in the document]
lamauny: Yes all pins can be configured as I2C.
If I remove I2C in json the defines `HAS_WIRE0 1` and `WIRE_INTERFACES_COUNT 1` are removed from board header, and then `LT_HW_I2C0` won't be defined.
Are these defines used ?

cores/lightning-ln882h/base/fixups/reg_cache.h
lamauny: unfortunately these reg files are included in some sdk header files which are needed in c++ modules...

kuba2k2: Well, then these SDK header files must be included with extern "C" in the C++ modules.

lamauny: extern "C" doesn't solve the issue, the problem is from a volatile variable assigned to a non-volatile local variable inside an inline function. C++ forbid this type of assignent and extern "C" changes nothing.
However I've found a smarter solution : in my `sdk_private.h` header i've added the definition of the sdk headers include guards and then these files are not parsed by the C++ compiler:
```
#ifdef __cplusplus
#define __REG_CACHE_H__
#define __REG_GPIO_H__
#define __REG_I2C_H__
#define __REG_LN_UART_H__
#define __REG_QSPI_H__
#define __REG_WDT_H__
#define __REG_SYSC_AWO_H__
#define __REG_SYSC_CMP_H__
extern "C" {
#endif // __cplusplus
```
What do you think ?

kuba2k2: Now, that's a much better explanation :slightly_smiling_face: 

As long as it works - this solution is fine by me. `sdk_private.h` is allowed to have SDK-specific fixes. Only make sure to add a short comment describing why that is needed (pretty much what you've described in the first two sentences of your reply).

boards/_base/ic/ln882hk.json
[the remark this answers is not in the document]
kuba2k2: Yes, these defines might be used somewhere.
Thanks for the clarification. The pin assignments have to stay then, at least for now, until a better solution is found.

cores/lightning-ln882h/base/api/lt_init.h
lamauny: I think if it works with others families it's a lucky break, there are a lot of discussions about this subject:
[how-to-make-gcc-link-strong-symbol-in-static-library-to-overwrite-weak-symbol](https://stackoverflow.com/questions/13089166/how-to-make-gcc-link-strong-symbol-in-static-library-to-overwrite-weak-symbol)
[gnu-ld-weak-declaration-overriding-strong-declaration](https://stackoverflow.com/questions/36994557/gnu-ld-weak-declaration-overriding-strong-declaration)
[override-weak-symbols-in-static-library](https://stackoverflow.com/questions/23079997/override-weak-symbols-in-static-library)
I've tried the proposed `-Wl,--whole-archive` linker option but this gives a lot of errors and gcc cannot link anymore.

lamauny: It's very strange because in fact I see no weak/strong definition of these functions other than mine, so I really don't understand why gcc doesn't link functions call with my implementation...
I would like some help because I'm stuck on this problem...

kuba2k2:
@lamauny 
Unfortunately, I'm unable to compile LibreTiny with this PR - it fails on Windows.
Because the OS is case-insensitive, the LN SDK is incorrectly picking up LT headers (in particular - `wifi.h` and `serial.h`) instead of the SDK headers. This causes inclusion of C++ code in C units, which fails for obvious reasons.

The SDK (on GitHub) will have to modified to rename the two header files (possibly more?). For instance, they could be prefixed with `ln_`.

cores/lightning-ln882h/base/api/lt_init.h
lamauny: I've found a fix by adding these linker options: `-Wl,--whole-archive -lcore_ln882h -Wl,--no-whole-archive`

lamauny:
> @lamauny Unfortunately, I'm unable to compile LibreTiny with this PR - it fails on Windows. Because the OS is case-insensitive, the LN SDK is incorrectly picking up LT headers (in particular - `wifi.h` and `serial.h`) instead of the SDK headers. This causes inclusion of C++ code in C units, which fails for obvious reasons.
> 
> The SDK (on GitHub) will have to modified to rename the two header files (possibly more?). For instance, they could be prefixed with `ln_`.

Could it be fixed by changing include path order ?

kuba2k2:
I don't think so. In that case LibreTiny would then pick up wrong headers.
The easiest way would be to rename the headers in SDK.

kuba2k2:
This solution would only work for the SDK .c files, but LibreTiny files would not be able to include the SDK's wifi.h when they are needed (in sdk_private.h).

lamauny:
maybe libretiny files could include `wifi_lib_export/wifi.h` when needed ?

kuba2k2:
I tested your idea, it won't work - LibreTiny's WiFi library is before fixups in the include path. The best solution (and the only possible one, I think) will be to rename the headers.

```
In file included from C:/Users/Kuba/.platformio/platforms/libretiny/cores/common/arduino/libraries/api/WiFi/WiFiEvents.h:9,
                 from C:/Users/Kuba/.platformio/platforms/libretiny/cores/common/arduino/src/Events.h:26,
(the one below is included incorrectly)
                 from C:/Users/Kuba/.platformio/platforms/libretiny/cores/common/arduino/libraries/api/WiFi/wifi.h:25,
                 from C:/Users/Kuba/.platformio/packages/framework-lightning-ln882h/components/utils/system_parameter.h:4,
                 from C:/Users/Kuba/.platformio/platforms/libretiny/cores/lightning-ln882h/base/sdk_private.h:56,
                 from C:/Users/Kuba/.platformio/platforms/libretiny/boards/variants/wl2s.c:6:
```

lamauny:
Is it wanted that fixups include paths are after common path ? just asking.

kuba2k2:
Not necessarily - but fixups are intended to replace SDK headers/code, not to solve complex problems with duplicated header names.

In the Beken SDK there was a similar issue, where FreeRTOS's `list.h` was named the same as Beken's `list.h` (two different headers). Simply renaming it was the easiest choice and works to this day.

lamauny:
@kuba2k2 do you think you can modify the SDK files and check the compilation on Windows platform ? I don't have a Windows machine on hand.

lamauny:
ltchiptool updated too

cores/lightning-ln882h/arduino/libraries/Serial/Serial.cpp
```diff
@@ -0,0 +1,99 @@
...
+	while (serial_read(serial_handles[2], &ch, 1)) {
+		data->buf.store_char(ch);
+	}
+}
+#else
+#define callback_uart2 NULL
+#endif
+
+// clang-format off
+const serial_rx_callbcak serial_rx_callbcaks[SER_PORT_NUM] = {
```
kuba2k2, line 48: Typo (yeah, I know it's like this in the SDK...):
```suggestion
const serial_rx_callbcak serial_rx_callbacks[SER_PORT_NUM] = {
```

cores/lightning-ln882h/arduino/libraries/Serial/Serial.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+	if (this->baudrate != baudrate || this->config != config)
+		this->configure(baudrate, config);
+}
+
+void SerialClass::configure(unsigned long baudrate, uint16_t config) {
+	if (!this->data)
+		return;
+
+	serial_init(serial_handles[port], (serial_port_id_t)port, baudrate, serial_rx_callbcaks[port]);
```
kuba2k2, line 69:
```suggestion
	serial_init(serial_handles[port], (serial_port_id_t)port, baudrate, serial_rx_callbacks[port]);
```

kuba2k2 (approved):

PR #313, opened 2025-01-11
[common] Make Serial RX buffer size configurable, increase default to 256 bytes

hn:
Increase RX buffer from 64 bytes to 256 bytes (ESPHome UART bus component default). Other sizes can easily be defined in ESPHome YAML e.g. like this:
```
bk72xx:
  board: cbu
  framework:
    options:
      LT_SERIAL_BUFFER_SIZE: 512
```

Preparatory work for https://github.com/esphome/issues/issues/6240. ESPHome could change `LT_SERIAL_BUFFER_SIZE` whenever `rx_buffer_size:` is changed.

Also covertly increases the send buffer for SoftwareSerial, but that should have no side effects.

### DIFF
--- a/cores/beken-72xx/arduino/libraries/Serial/SerialPrivate.h
+++ b/cores/beken-72xx/arduino/libraries/Serial/SerialPrivate.h
@@ -6,9 +6,9 @@
 #include <sdk_private.h>
 
 typedef struct {
-	RingBuffer buf;
+	SerialRingBuffer buf;
 } SerialData;
 
 #define DATA ((SerialData *)data)
 #define BUF	 (DATA->buf)
-#define pBUF ((RingBuffer *)param)
+#define pBUF ((SerialRingBuffer *)param)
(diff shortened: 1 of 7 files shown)


hn:
Basic tests done with `cbu` and `rtl8710bx` modules.

hn:
Withdraw the above statement: `rx_buffer_size:` is a per-device (per-UART) setting, while `LT_SERIAL_BUFFER_SIZE ` would be a global setting. That would be somewhat inconsistent.

Anyway, it is an advantage if the standard buffer size is increased and made adjustable.

kuba2k2:
I know it's a per-UART setting, but it's the most logical place to put it. So unless `rx_buffer_size:` controls ESPHome's internal buffer, it should control LibreTiny's global buffer. If there is more than one UART component used (likely never), prohibit changing the buffer size on one of them only.

kuba2k2 (approved):

PR #321, opened 2025-06-24
Don't re-define c++ round() method

szupi-ipuzs:
I've recently been hit by the same compilation error as described [here](https://github.com/esphome/issues/issues/6124).
The easiest way to solve this is to only define this macro if the header is included from within pure C code.

Works for me :)


szupi-ipuzs:
Sorry the linked issue is not the correct one. Now I can't find it again...

PR #332, opened 2025-09-12
[boards] Add WB02A board

unusual-thoughts:
another LN882H variant I found in a smart switch
wb02a

### DIFF
--- a/boards/_base/pcb/wb02a-test.json
+++ b/boards/_base/pcb/wb02a-test.json
@@ -1,0 +1,93 @@
+{
+	"pcb": {
+		"test_pads": {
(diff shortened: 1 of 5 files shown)


divadiow:
may I enquire as to which device exactly this was found and if you took a backup of the flash?

unusual-thoughts:
It's this one: https://fr.aliexpress.com/item/1005007098797755.html

Unfortunately I wiped the firmware before I was able to back it up, but both OpenBeken and esphome work fine

The creator of this flashing tool https://github.com/mandl/LN882Loader apparently has the same board too (see pictures)

kuba2k2 (approved):

PR #343, opened 2026-01-08
[common] Automatically cleanup OTA context in begin()

szupi-ipuzs:
In case OTA is interrupted by the user, the ota-handling code in esphome does not call end(). Instead it will immediately call begin() again if the user restarts the ota, which would fail because of that. The esphome team insist that this is correct behavior on their behalf and libretiny ota backend needs to handle such cases gracefully.


szupi-ipuzs:
Right, sorry about that. I didn't want to cherry-pick from my private branch so i edited the file in-place and made this mistake :D

PR #359, opened 2026-02-06
ambz2 uart handling using serial_api

szupi-ipuzs:
The sdk examples for serial show that critical section should be entered both when installing and reinstalling the hook.


szupi-ipuzs:
Ok, eventually this change went in the other direction - it reimplements the ambz2 uart functionality using the serial_api functions from the SDK. I think it's safer to use them instead of trying to guess which lower-level methods to call / bits to set.

I also had to make sure that once logging uart is disabled, then lt_log() don't try to write to it, otherwise hangups occur.

UART2 can be reused as a normal uart if needed, but this requires to call sys_log_uart_off() before.
It is also important to note that the system does not restore the syslog logger automatically when soft reboot occurs, therefore the call to sys_log_uart_on() had to be added to lt_reset().

Tested this so far with UART2 (PA15/PA16) as logger and UART0 (PA13/PA14) as tuyamcu uart. More tests coming.

kuba2k2 (changes requested):

cores/realtek-ambz2/arduino/libraries/Serial/Serial.cpp
```diff
@@ -3,89 +3,127 @@
...
+	serial_init(this->data->uart, static_cast<PinName>(this->tx), static_cast<PinName>(this->rx));
 }
 
 void SerialClass::configure(unsigned long baudrate, uint16_t config) {
 	if (!this->data)
 		return;
 
-	uint8_t dataWidth = (config & SERIAL_DATA_MASK) == SERIAL_DATA_7 ? 7 : 8;
-	uint8_t parity	  = (config & SERIAL_PARITY_MASK) ^ 0b11;
-	uint8_t stopBits  = (config & SERIAL_STOP_BIT_MASK) == SERIAL_STOP_BIT_2 ? 2 : 1;
```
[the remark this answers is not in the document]
szupi-ipuzs: Actually I've now realized the previous version had a bug. See the actual values of the parity constants between arduino:
image


and realtek sdk:

image


This means direct casting should've never been used. That's why using function is always better :) And even though they increase the line count (what's wrong with that?), they usually get inlined by the compiler anyway. So the readability improves, but the code size doesn't actually increase.

szupi-ipuzs:
Ok, take your time then :)
In the meantime, there's one more thing that i've noticed: in the scenarios where UART2 is reconfigured to be the tuyamcu uart and not logger, I can still see the realtek drivers (eg. the wifi driver) to actually use it for printing info. This will screw the tuyamcu communication up. I wonder if there's a way to tell the drivers to use a different uart. Or to not use it at all?

cores/realtek-amb/base/sdk_private.h
```diff
@@ -24,6 +24,8 @@ extern "C" {
 #include <hal.h>
 #include <hal_sys_ctrl.h>
 #include <rtl8710c.h>
+#include <serial_api.h>
+#include <serial_ex_api.h>
```
kuba2k2, line 28: Move to mbed API includes below - this isn't chip-specific. In fact, `serial_api.h` is already included there.

cores/realtek-ambz2/arduino/libraries/Serial/Serial.cpp
```diff
@@ -4,88 +4,124 @@
...
+
+static uint8_t datawidth_from_config(const uint16_t config) {
+	return (config & SERIAL_DATA_MASK) == SERIAL_DATA_7 ? 7 : 8;
+}
+
+static uint8_t stopbits_from_config(const uint16_t config) {
+	return (config & SERIAL_STOP_BIT_MASK) == SERIAL_STOP_BIT_2 ? 2 : 1;
+}
+
+static void irq_callback(SerialData *sd, int event) {
```
kuba2k2, line 36: This is called `callback()` and has a `*data` parameter on other platforms, why not keep the convention here too?

cores/realtek-ambz2/arduino/libraries/Serial/Serial.cpp
```diff
@@ -4,88 +4,124 @@
...
 	if (event != RxIrq)
 		return;
-	hal_uart_adapter_t *uart = data->uart;
 
-	uint8_t c;
-	while (hal_uart_rgetc(uart, (char *)&c)) {
+	uint8_t c = serial_getc(sd->uart);
```
kuba2k2, line 40: The previous implementation with the `while()` loop had an advantage of reading all available characters during a single interrupt. The current solution only reads a single character at a time, which could possibly introduce unwanted delays.

The `hal_uart_rgetc()` function can still be used here, as `serial_t` (`sd->uart`) has a member called `->uart_adp`, which is of type `hal_uart_adapter_t`.

cores/realtek-ambz2/arduino/libraries/Serial/Serial.cpp
```diff
@@ -4,88 +4,124 @@
...
 
 	this->baudrate = baudrate;
 	this->config   = config;
 }
 
 void SerialClass::endPrivate() {
+	disable_syslog_on_port(this->port);
+	if (lt_log_get_port() == this->port) {
+		lt_log_disable();
+	}
```
kuba2k2, line 95: This still doesn't seem right to me.

The `Serial` class shouldn't really touch the LT logger, nor the SDK logger for that matter. With this solution, one can indeed disable logging completely using `Serial.end()`, but can't enable it back using just `Serial.begin()`.

If there ever is a device using UART2 (logger) for something else than logging, then perhaps the user should just change the default output port to something else, in their own config?

cores/realtek-ambz2/base/port/printf.c
```diff
@@ -12,13 +12,16 @@ static UART0_Type *uart_dev[4] = {
...
+extern uint32_t lt_uart_port;
 
 void putchar_(char c) {
 	putchar_p(c, lt_uart_port);
 }
 
 void putchar_p(char c, unsigned long port) {
+	if (port >= 4) {
+		return;
+	}
```
kuba2k2, line 24: Seems like an unnecessary slowdown. No sane piece of code will call putchar with port 4...

cores/realtek-ambz2/base/port/printf.c
```diff
@@ -12,13 +12,16 @@ static UART0_Type *uart_dev[4] = {
 	UART3,
 };
 
-uint8_t lt_uart_port = 2;
+extern uint32_t lt_uart_port;
```
kuba2k2, line 15: This ties both SDK and LT logger ports together and forever - contrary to the documentation:

> - LT_UART_DEFAULT_PORT (unset) - default output port for all messages (SDK, LT logger, Serial class); can be 0, 1 or 2
> - LT_UART_DEFAULT_LOGGER (unset) - override default output port for LT logger only

The `putchar_` function is used by `printf()`, which in turn is used by the SDK. It is initialized in `lt_init_family()` to `LT_UART_DEFAULT_PORT`, so that controls SDK output.

LT logger has *its own* output port setting, which can be changed using `lt_log_set_port()`, and which *defaults* to `LT_UART_DEFAULT_LOGGER`.

The output port of pretty much everything logging-related at once can be changed on compile-time. The logger has its own setting which can also be changed on runtime; that possibility was introduced for... not sure what reason exactly, but it shouldn't be changed in the current major version anyway.

---

TL;DR, I'm just not sure what this change should help to accomplish.

kuba2k2 (changes requested):

PR #362, opened 2026-02-09
Remove outdated ESPHome port information

szupi-ipuzs:
I think it's about time someone changes it :)


kuba2k2:
Well, if we're already removing outdated info, then I don't think that will suffice:
- there's an ESPHome page in LT docs [here](https://docs.libretiny.eu/docs/projects/esphome/), it doesn't really have anything meaningful and has outdated YAML examples which don't work anymore,
- the sidebar links to a [pre-merge HA addon](https://github.com/libretiny-eu/esphome-hass-addon/pkgs/container/libretiny-esphome-hassio),
- the [Getting Started](https://docs.libretiny.eu/docs/getting-started/) page also mentions the old ESPHome port.

Instead, I'd probably remove the HA addon completely from the sidebar. The "ESPHome setup guide" should probably link to ESPHome's documentation, same as the entry in the main README.

PR #364, opened 2026-02-11
[ln788h] Make sure printfs get redirected to correct uart

szupi-ipuzs:
Same problem as on ambz2 - the printfs must always write to whichever uart is set as logger, not to the default one.

### DIFF
--- a/cores/lightning-ln882h/base/port/printf.c
+++ b/cores/lightning-ln882h/base/port/printf.c
@@ -7,13 +7,16 @@
 
 extern Serial_t *serial_handles[SER_PORT_NUM];
 
-uint8_t uart_print_port = LT_UART_DEFAULT_LOGGER;
+extern uint32_t lt_uart_port;
 
 void putchar_(char c) {
-	putchar_p(c, uart_print_port);
+	putchar_p(c, lt_uart_port);
 }
 
 void putchar_p(char c, unsigned long port) {
+	if (port >= SER_PORT_NUM) {
+		return;
+	}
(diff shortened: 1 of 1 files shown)


Bl00d-B0b:
## Analysis

This PR has two independent changes:

1. **Bounds check** — `if (port >= SER_PORT_NUM) return;` in `putchar_p()`. Straightforward safety guard, no issues.

2. **Variable rename** — replaces the `uart_print_port` local variable with `extern uint32_t lt_uart_port`.

### Problem with the rename

`lt_uart_port` is referenced as `extern` but **is not defined anywhere in the LN882H family**. On ambz2 it is defined in that family's own `printf.c` (`uint8_t lt_uart_port = 2;`), but no equivalent definition exists for LN882H. As-is this would produce a linker error.

### Conflict with #373

PR #373 (now merged) restructured the LN882H logger setup around `uart_print_port`:
- `printf.c` owns the definition: `uint8_t uart_print_port = LT_UART_DEFAULT_LOGGER;`
- `lt_init.c` holds `extern uint8_t uart_print_port;` and sets it to UART1 in `lt_init_log()`

If this PR merges, `uart_print_port` disappears from `printf.c` while `lt_init.c` still references it — compile/link failure.

### What a correct fix would look like

To align LN882H with the ambz2 pattern, `lt_uart_port` needs to be **defined** (not just declared) in LN882H, and `lt_init.c` updated to set `lt_uart_port` instead of `uart_print_port`. The bounds check can land independently and cleanly without any of this.

PR #368, opened 2026-03-03
[realtek-amb] Reduce main task stack from 16KB to 8KB

bdraco:
The main task stack is set to `4096 * 4` = 16384 bytes. Runtime profiling with FreeRTOS stack watermark scanning shows only ~1.3KB is actually used, wasting ~8KB of RAM on all Realtek families (amb, ambz, ambz2) that inherit this startup code.

Reduce to 8192 bytes, which matches BK72xx and LN882H, and still leaves ~6.4KB of headroom. This is the same value ESPHome uses for ESP32


bdraco:
Thanks

PR #372, opened 2026-04-02
[boards] Add Tuya WL2H-U and LN-CB3S (LN882H)

Bl00d-B0b:
## Summary

Add board definitions for two LN882H-based modules:

**Tuya WL2H-U** — standard Tuya Wi-Fi/BLE module (15.8×20.3mm, 21 pads)
- `boards/wl2h-u.json`
- `boards/_base/pcb/wl2h-u.json` — physical pinout using `tuyau` template
- `boards/variants/wl2h-u.c/.h` — Arduino pin map (19 GPIOs: PA0–PA7, PA9–PA12, PB3–PB9)

**LN-CB3S V1.0** — custom CB3S-footprint module (16×24mm, 22 pads) carrying WL2H-U-2 (LN882H), found in DS-101JL / BSEED wall switches
- `boards/ln-cb3s-v1.0.json`
- `boards/_base/pcb/ln-cb3s-v1.0.json` — CB3S-compatible PCB layout (`esp12e-22`/`tuya-16x24`)
- `boards/variants/ln-cb3s-v1.0.c/.h` — Arduino pin map (12 GPIOs: PA0–PA6, PA9, PB4–PB6, PB9)

### Hardware specs (WL2H-U / LN-CB3S V1.0)
- **CPU**: ARM Cortex-M4F @ 160 MHz
- **RAM**: 296 KB
- **Flash**: 2 MB
- **Wi-Fi**: 802.11b/g/n
- **BLE**: 5.1
- **FCC ID** (WL2H-U): 2ANDL-WL2H-U

### Notes
- WL2H-U pinout verified against the [official Tuya datasheet](https://developer.tuya.com/en/docs/iot/WL2H-U-Module-Datasheet?id=Kbohlj8eg19u5)
- LN-CB3S V1.0 pinout derived from community schematic (elektroda.com topic4162022)
- Variant files were hand-generated; please regenerate with boardgen if needed

PR #373, opened 2026-04-02
[lightning-ln882h] Change default log output port to UART1

Bl00d-B0b:
## Problem

On LN882H, `lt_init_log()` unconditionally calls `serial_init()` during startup, which triggers `uart_io_pin_request()` → sets the AFIO pin mux for the default UART port's TX/RX pins. This permanently claims those pins (e.g. PA2/PA3 for UART0) **even when `logger: baud_rate: 0` is set in ESPHome**, preventing them from being used as GPIO.

On BK72XX this is not an issue because UART0 is pre-initialized by the Beken bootloader — LibreTiny never claims it explicitly.

## Root cause

```c
// lt_init.c — always runs, regardless of logger settings
static void lt_init_log(void) {
    uart_print_port = LT_UART_DEFAULT_LOGGER;
    serial_init(&m_LogSerial, LT_UART_DEFAULT_PORT, CFG_UART_BAUDRATE_LOG, NULL);
    //           ^^^ calls uart_io_pin_request() → AFIO claims PA2/PA3
}
```

## Fix

When `LT_UART_SILENT_ALL=1` (set by ESPHome's `sdk_silent: all`, which is the **default**), all SDK UART output is already suppressed via `printf_config.h` — the logger serial is never written to. Skipping `serial_init()` in this case is safe and frees the UART pins for GPIO use.

When serial output is actually needed, ESPHome's logger component or an explicit `Serial.begin()` call initializes the hardware instead.

## Effect

With this fix, setting `logger: baud_rate: 0` in ESPHome (combined with the default `sdk_silent: all`) frees the default UART TX/RX pins for GPIO use on LN882H — consistent with BK72XX behaviour.

Without the fix, users must work around this by setting `LT_UART_DEFAULT_PORT: 1` in their ESPHome config to redirect the logger to a different UART port.

### DIFF
--- a/cores/lightning-ln882h/base/lt_family.h
+++ b/cores/lightning-ln882h/base/lt_family.h
@@ -5,11 +5,13 @@
 #include <lt_pins.h>
 
 // Choose the main UART output port
+// Per Tuya WL2H-U datasheet, UART1 (PB8/PB9) is the designated log port;
+// UART0 (PA2/PA3) is the user UART and should remain free as GPIO.
 #ifndef LT_UART_DEFAULT_PORT
-#if LT_HW_UART0
-#define LT_UART_DEFAULT_PORT 0
-#elif LT_HW_UART1
+#if LT_HW_UART1
 #define LT_UART_DEFAULT_PORT 1
+#elif LT_HW_UART0
+#define LT_UART_DEFAULT_PORT 0
 #elif LT_HW_UART2
 #define LT_UART_DEFAULT_PORT 2
 #else


kuba2k2:
The `LT_UART_SILENT_ALL` doesn't apply to LT logger, but only to vendor SDK logging (using `printf()`). Initializing the UART shouldn't be disabled based on that configuration option, because it serves a different purpose. See: [documentation](https://docs.libretiny.eu/docs/dev/config/#serial-output).

A better solution would be to use the `LT_LOGGER` flag. This allows to disable the LT logger output entirely, but currently the UART is still initialized. Making the initialization conditional on this flag would indeed free the UART TX/RX pins from the logger.

There is however another component that uses UART internally in LT - [Auto-download-reboot](https://docs.libretiny.eu/docs/flashing/tools/adr/). When it's enabled (the default option), it listens for flashing commands on the default flashing port (depends on the chip). To do that, it needs to initialize the UART. The `LT_AUTO_DOWNLOAD_REBOOT` flag is enabled by default, but it's currently not implemented on LN882, so it doesn't make a difference anyway.

---

Note that, on most off-the-shelf devices, the UART used for log output is usually not connected to any other peripherals. Because of this I don't really see the purpose of using these pins for anything else, unless you're using some kind of development board, or have unsoldered the module and use it as a bare module.

Bl00d-B0b:
Thanks for the detailed explanation, that makes sense.

**Real-world context:** Just to address the skepticism — this is not a dev board or bare module. The device is a DS-101JL dual-gang wall switch. The MCU module is an **LN-CB3S V1.0** — a pin-compatible drop-in replacement for the Tuya CB3S (Beken BK7231N) module, but carrying a **WL2H-U-2** chip (Tuya's Lightning Semi LN882H variant) instead. Because it reuses the CB3S footprint on a mass-produced PCB, PA2 is physically connected to the right indicator LED on the switch.

LN-CB3S V1.0 board with WL2H-U-2 module

---

**Reworked:** I've updated the PR to use the correct fix — changing `lt_family.h` to prefer UART1 over UART0 as the default log port, consistent with how `beken-72xx` avoids UART0. The `lt_init.c` conditional has been reverted entirely. This matches the Tuya WL2H-U datasheet and fixes the root cause introduced in #312 for all LN882H boards, not just boards with `LT_UART_SILENT_ALL`.

kuba2k2:
> when ESPHome's logger: baud_rate: 0 is configured, does that currently map to LT_LOGGER=0 in LibreTiny? 

No, it doesn't. `baud_rate: 0` only causes ESPHome to initialize the `Serial` instance (which is already initialized by the logger), and to disable the `Serial.println()` operations completely.

> My understanding is it sets LT_UART_SILENT_ALL=1 [...]

It doesn't either. This is a separate option, controlled by ESPHome's `sdk_silent:` option.

> would a combined condition like !LT_LOGGER || LT_UART_DEFAULT_PORT != 0 be more appropriate 

`LT_UART_DEFAULT_PORT` *can* have a valid value of 0 - this is the UART index, which on LN882x can be UART0, 1 or 2 (depending on available board pins).

---

Disabling the UART initialization on `!LT_LOGGER` is definitely a valid fix to implement, but it won't fix your issues entirely.

Your device's LED is connected to GPIOA2, which is also the UART0 TX pin. On the LN-CB3S module, it's available on pad 16 - which corresponds to all other CB3S-like modules (or even ESP12F modules). On all prior Beken-based boards this port was NOT used for the logging output.

Per [Tuya documentation on WL2H-U](https://developer.tuya.com/en/docs/iot/WL2H-U-Module-Datasheet?id=Kbohlj8eg19u5#title-6-Pin%20definition), this also applies to LN882x modules - the correct logging output port is UART1, not UART0.

Your device was designed with this in mind - the original firmware was most definitely using UART1, so the UART0 port was free to use as the LED indicator.

---

The real solution would probably be to change the default logging port of LibreTiny to UART1 - we usually try to stick to the manufacturer-defined UART usage conventions. The LN882H support was added in PR #312, and apparently it was just implemented incorrectly for this whole time.

Bl00d-B0b:
Thanks for digging into the root cause — you're right, the correct fix is much simpler. I've reworked the PR: `lt_family.h` now prefers UART1 over UART0 as the default log port (mirroring how `beken-72xx` avoids UART0), and the `lt_init.c` conditional has been reverted entirely. This fixes the issue at the source for all LN882H boards rather than working around it per-config.

kuba2k2 (approved):

Bl00d-B0b:
Hi @kuba2k2 — just a gentle ping on this PR and the related ones I opened earlier this month:

- #372 [boards] Add Tuya WL2H-U (LN882H)
- #373 [lightning-ln882h] Change default log output port to UART1
- #375 [lightning-ln882h] Fix static IP ignored after WiFi connect
- #376 [lightning-ln882h] Fix -Wregister warnings from CmBacktrace
- #378 [boards] Add Tuya T34 (BK7231N, LGA36)

Happy to rebase or adjust anything if that helps with review. Thanks for the project!

PR #375, opened 2026-04-02
[lightning-ln882h] Add static IP support

Bl00d-B0b:
Adds static IP support for LN882H. The LN882H SDK always starts DHCP internally in \wifi_sta_connect()\ with no option to disable it, so static IP is not supported at the SDK level. This PR implements static IP by stopping DHCP and applying the configured address after each connect, reconnect and roam event.

### Changes

1. **Apply static IP after connect** — stop DHCP and apply static IP in \wifiEventStaConnected()\ before the SDK hands over to the application. \wifi_sta_connect()\ always starts DHCP internally with no way to disable it.

2. **Fix forward declaration of \wifiEventIpReceived\** — required for the above.

3. **Ensure static IP survives reconnect and roam** — two additional cases:
   - Wrong order: \
etif_set_addr()\ was called before \
etifapi_dhcp_release_and_stop()\. When DHCP is already BOUND, this triggers the netif status callback with \dhcp_supplied_address==true\ before DHCP is stopped. Fixed: stop DHCP first.
   - Missing fallback: \wifiEventIpReceived()\ is called from inside \dhcp_recv()\ in the LwIP tcpip_thread. If DHCP completes despite the early stop (roaming, SDK-internal reconnect), the lease stays active. Cannot call \
etifapi_dhcp_release_and_stop()\ (deadlock) or \dhcp_release_and_stop()\ directly (use-after-free). Fixed: schedule \wifiApplyStaticIpCallback\ via \	cpip_callback()\ to run after \dhcp_recv()\ returns.

### Tested on
LN882H-based 2-gang light switch (DS-101JL) running ESPHome — static IP stable across power cycles, reconnects and roaming for several months.

### DIFF
--- a/builder/family/lightning-ln882h.py
+++ b/builder/family/lightning-ln882h.py
@@ -32,7 +32,9 @@ queue.AppendPublic(
         "-Wno-write-strings",
         "-Wno-maybe-uninitialized",
     ],
-    CXXFLAGS=[],
+    CXXFLAGS=[
+        "-Wno-register",  # cmb_def.h uses 'register' keyword (removed in C++17)
+    ],
     CPPDEFINES=[
         # other options
         "ARM_MATH_CM4",
--- a/cores/lightning-ln882h/arduino/libraries/WiFi/WiFiEvents.cpp
+++ b/cores/lightning-ln882h/arduino/libraries/WiFi/WiFiEvents.cpp
@@ -28,6 +28,22 @@ static void wifiEventStaConnected(void *arg) {
 	memcpy(eventInfo.wifi_sta_connected.bssid, pWiFi->BSSID(), 6);
 
 	pWiFi->postEvent(ARDUINO_EVENT_WIFI_STA_CONNECTED, eventInfo);
+
+	// If static IP is configured, apply it now and stop DHCP before the SDK
+	// starts it. wifi_sta_connect() always starts DHCP internally; we have no
+	// way to pass dhcp_mode to it (unlike BK72XX's bk_wlan_start_sta_adv_fix).
+	WiFiNetworkInfo &info = pDATA->sta;
+	if (info.localIP) {
+		struct netif *ifs = netdev_get_netif(NETIF_IDX_STA);
+		ip4_addr_t ipaddr, netmask, gw;
+		ipaddr.addr	 = info.localIP;
+		netmask.addr = info.subnet;
+		gw.addr		 = info.gateway;
+		netif_set_addr(ifs, &ipaddr, &netmask, &gw);
+		netifapi_dhcp_release_and_stop(ifs);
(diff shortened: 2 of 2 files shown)


Bl00d-B0b:
Added one more fix: on the reconnect/roam fallback path (`wifiApplyStaticIpCallback`), LwIP's `dhcp_recv()` processes DHCP option 6 and overwrites the DNS table *before* we get to stop DHCP. The static IP and gateway were already being restored, but DNS was not. Added DNS restoration (`dns_setserver`) after `netif_set_addr()` using the stored `info.dns1`/`info.dns2` values.

Currently testing on a **DS-101JL / BSEED WiFi Touch Light Switch 2 Gang** with an **LN-CB3S V1.0** module (WL2H-U-2 / LN882H).

Bl00d-B0b:
## Crash with ESPHome 2026.5.0

Tested this branch (`6bf1b38`) against ESPHome 2026.5.0 — the device (DS-101JL / generic-ln882hki) enters a crash loop on every boot. The crash happens **before WiFi association completes** — the device is not pingable and never opens OTA port 8820.

**Key finding:** the branch is already based on top of v1.12.1 (0 commits behind). Default LibreTiny v1.12.1 works fine with ESPHome 2026.5.0. The crash is therefore specific to the `WiFiEvents.cpp` changes in this PR.

**Suspected cause:** `wifiEventStaConnected()` now calls `wifiEventIpReceived()` directly, which posts `ARDUINO_EVENT_WIFI_STA_GOT_IP` and triggers ESPHome's component initialisation chain. Something in that chain changed in ESPHome 2026.5.0 and causes a hard fault before the network interface is fully up — which is why the device never appears on the network at all.

No UART crash log captured yet (devices are currently unrecoverable over OTA; serial flash + UART session needed to get a stack trace). Will follow up with the exact fault address once that is done.

Workaround for now: comment out the `source:` line and use default LibreTiny v1.12.1 (loses static IP, use router DHCP reservation instead).

kuba2k2:
There is only one pathway to starting DHCP internally by the SDK:
1. `sta_connected()` in wifi_manager.c
2. `netdev_set_state()` in ethernetif.c
3. Finally, `netifapi_dhcp_start()` in netifapi.h

The wifi_manager will always call `event_cb[WIFI_MGR_EVENT_STA_CONNECTED]()` when the station is connected. Because of `wifi_manager_reg_event_callback(WIFI_MGR_EVENT_STA_CONNECTED, &wifiEventStaConnected);`, the WiFi library function will be called, and only *after* DHCP has been started internally. Note: it is **not** started by `wifi_sta_connect()`, in contrary to the PR's source code comments.

So only the part that sets a static IP address in STA_CONNECTED should be needed.

Tested on ESPHome, works just fine; forcefully kicking the device off the network causes it to reconnect and use the static IP address again.

PR #376, opened 2026-04-02
[lightning-ln882h] Fix -Wregister warnings from CmBacktrace cmb_def.h

Bl00d-B0b:
The LN882H SDK CmBacktrace library (`cmb_def.h`) uses the `register` storage class specifier in three inline assembly helpers: `cmb_get_msp()`, `cmb_get_psp()`, and `cmb_get_sp()`.

This keyword was removed in C++17 (P0001R1), producing spurious `-Wregister` warnings during every build when compiled with `-std=c++17` (as ESPHome does):

```
cmb_def.h:408:27: warning: ISO C++17 does not allow register storage class specifier [-Wregister]
  408 |         register uint32_t result;
```

The warning appears 9 times per build (3 inline functions x 3 translation units that include the header).

**Fix:** Add `-Wno-register` to `CXXFLAGS` in `builder/family/lightning-ln882h.py`. This is a no-op for plain LibreTiny builds (which do not use C++17) and only suppresses the warning for C++ consumers such as ESPHome.

No SDK sources are modified.

### DIFF
--- a/builder/family/lightning-ln882h.py
+++ b/builder/family/lightning-ln882h.py
@@ -32,7 +32,9 @@ queue.AppendPublic(
         "-Wno-write-strings",
         "-Wno-maybe-uninitialized",
     ],
-    CXXFLAGS=[],
+    CXXFLAGS=[
+        "-Wno-register",  # cmb_def.h uses 'register' keyword (removed in C++17)
+    ],
     CPPDEFINES=[
         # other options
         "ARM_MATH_CM4",


kuba2k2:
Why not just add `-Wno-register` to GCC compile flags for the entire build?

Side note, LibreTiny doesn't use C++17 by default; ESPHome does.

Bl00d-B0b:
Agreed, `-Wno-register` is simpler and cleaner. Happy to rework the PR to use that instead — would just need to revert the `sdk_private.h` change and add the flag in the right place.

Which build file should it go in? I am thinking somewhere in the `lightning-ln882h` platform build environment (e.g. a SConscript or platform builder), but I am not familiar enough with the LibreTiny build system to know the canonical location for per-platform compiler flags.

kuba2k2:
The most logical place to add it would be the file where all other options are defined, `builder/family/lightning-ln882h.py`. Note that this warning only concerns C++ code, so it should go to the `CXXFLAGS` variable, instead of the `CFLAGS`.

Bl00d-B0b:
Done — reworked the PR to use `-Wno-register` in `CXXFLAGS` in `builder/family/lightning-ln882h.py` as you suggested, and the `sdk_private.h` pragma change has been reverted.

kuba2k2 (approved):

PR #381, opened 2026-05-24
[lightning-ln882h] Fix duplicate MAC addresses on LN882H (fixes #338)

### LINKED ISSUE #338: LN882HKI  - MAC address not recognized
I have two Ali-Express power monitoring smart plugs using the WL2S WiFi chip. I have flashed both with ESPHome (2025.11.01-dev), using the manufacturer's LN882H_CMD_TOOL (ltchiptool fails). 

The devices appear on the home network with the SAME MAC address and they do not accept manual IP assignment.

The MAC access is always: 00:50:C2:5E:10:88, for both devices.

I decided to use the LN882H_CMD_TOOL to read the MAC address: It returns FF:FF:FF:FF:FF:FF

I decided to use the LN882H_CMD_TOOL to burn unique MAC addresses to both chips. Although the tool confirms the correct address, they still appear under the "original" address in the WiFi network, so I can's have both devices on the network at the same time.

It appears LibreTiny and/or ESPHome is not recognizing the MAC address in the chip at all.

Bl00d-B0b:
Fixes #338

## Problem

Every LN882H device ships with the same factory-default MAC address (`00:50:C2:5E:10:88`) stored in KV flash. `sysparam_integrity_check_all()` (called during `lt_init_family()`) writes this placeholder on first boot and never replaces it. Every device on the same network therefore shares one MAC, causing ARP conflicts and connection failures.

## Fix

Add `lt_init_unique_mac()` called immediately after `sysparam_integrity_check_all()` in `lt_init.c`. It resolves the MAC address in priority order — once, on first boot after flash. Every subsequent boot returns in step 1 immediately.

### Priority order

**Step 1 — sysparam KV already unique**
Read the stored STA MAC via `sysparam_sta_mac_get()`. If it differs from the factory sentinel, return immediately. This is the fast path on every boot after the first, and fully covers the LibreTiny→LibreTiny OTA case: the sysparam KV lives at `0x1E0000`, which is outside the OTA partition and is never erased by OTA.

**Step 2 — Restore Tuya factory MAC from BLK_V1.0 KV**
Tuya assigns every LN882H a unique MAC at the factory and writes it as a second `6_sta_mac` entry in a BLK_V1.0 append-log KV store at flash offset `0x1C5000`. Scan for the last valid non-sentinel entry and restore it to sysparam KV.

This region lies within the OTA partition (`0x133000`–`0x1DD000`), but **no partition layout changes are needed**: OTA writes are block-addressed and the current firmware (~530 KB) ends at ~`0x1B4700`, well short of `0x1C5000`. UART full-flash similarly only covers `0x000000`–`0x088000`. The Tuya KV is naturally preserved under both flash methods.

**Step 3 — TRNG fallback**
If the Tuya KV is absent or erased (e.g. UART flash with full-chip erase option), generate a unique address via `ln_generate_random_mac()` (hardware TRNG, already compiled in via `ln882h_net`). This mirrors the approach used in OpenBeken.

In all cases the SoftAP MAC is derived by setting the locally-administered bit (`mac[0] |= 0x02`) on the STA MAC.

## Analysis

This fix is based on binary examination of **5 unique Tuya LN882H flash dumps** (4× DS-101JL wall switch + 1× GU10 bulb from issue #338), including devices in three states: never provisioned, provisioned via Tuya app, and provisioned and in use for weeks.

Key findings:
- The unique MAC is **factory-assigned at manufacturing time**, not during Tuya app provisioning — even devices that were never connected to a Tuya account carry a unique `6_sta_mac` entry
- The BLK_V1.0 KV region (`0x1C5000`–`0x1C9000`, 4 blocks × 4 KB) has an identical layout across all examined devices: sentinel at `0x1C51EB`, unique MAC at `0x1C5423`
- BLK_V1.0 keys are **not** null-terminated; the format is `it_magic(8) + crc16(2) + val_size_u16_le(2) + prev_ptr_u32_le(4) + key_raw(N) + value_raw`
- LibreTiny's OTA writer (`uf2_write`) uses `offset = part->offset + block->addr` — it only erases sectors for UF2 blocks actually present in the image; no block targets `0x1C5000`

Detailed analysis is posted as a comment on this PR.

## Safety

- `lt_tuya_kv_read_sta_mac()` is read-only; it never writes to flash
- `ln_generate_random_mac()` is not thread-safe but is called before the RTOS scheduler starts
- The sentinel check (`memcmp` against `{0x00,0x50,0xC2,0x5E,0x10,0x88}`) ensures existing devices with a persisted unique MAC in sysparam KV are never affected

### DIFF
--- a/boards/_base/lightning-ln882hki-tuya.json
+++ b/boards/_base/lightning-ln882hki-tuya.json
@@ -1,0 +1,10 @@
+{
+	"flash": {
+		"tuya_kv": "0x1C5000+0x4000"
+	},
+	"doc": {
+		"params": {
+			"layout": "TuyaOpen LN882H default: KV partition starts at 0x001C5000 and size is 16 KB / 0x4000"
+		}
+	}
+}
--- a/boards/generic-ln882hki-tuya.json
+++ b/boards/generic-ln882hki-tuya.json
@@ -1,0 +1,16 @@
+{
+	"_base": [
+		"generic",
+		"lightning-ln882x",
+		"lightning-ln882hki",
+		"lightning-ln882hki-tuya",
+		"ic/ln882hk"
+	],
+	"build": {
+		"mcu": "ln882hk",
+		"variant": "generic-ln882hki-tuya"
+	},
+	"name": "Generic - LN882HKI (Tuya)",
+	"symbol": "LN882HKI_TUYA",
+	"vendor": "Tuya"
+}
--- a/boards/ln-cb3s-v1.0.json
+++ b/boards/ln-cb3s-v1.0.json
@@ -2,6 +2,7 @@
 	"_base": [
 		"lightning-ln882x",
 		"lightning-ln882hki",
+		"lightning-ln882hki-tuya",
 		"ic/ln882hk",
 		"pcb/ln-cb3s-v1.0"
 	],
--- a/cores/lightning-ln882h/base/api/lt_init.c
+++ b/cores/lightning-ln882h/base/api/lt_init.c
@@ -3,6 +3,19 @@
 #include <libretiny.h>
 #include <sdk_private.h>
 
+// Declared in components/utils/ln_misc.h; compiled in ln882h_net.
+// Not thread-safe - safe here since lt_init_family() runs before the RTOS scheduler.
+extern int ln_generate_random_mac(uint8_t *addr);
+
+// hal_flash_read: only needed on Tuya board variants (FLASH_TUYA_KV_OFFSET defined).
+// Forward-declared to avoid the hexdump macro conflict in hal_flash.h.
+#ifdef FLASH_TUYA_KV_OFFSET
+extern uint8_t hal_flash_read(uint32_t offset, uint32_t length, uint8_t *buffer);
+#endif
+
+// Factory-default STA MAC on every LN882H - sentinel for "not yet unique".
+static const uint8_t WIFI_MAC_FACTORY_DEFAULT[6] = {0x00, 0x50, 0xC2, 0x5E, 0x10, 0x88};
+
 extern uint8_t uart_print_port;
 extern Serial_t m_LogSerial;
 
@@ -11,6 +24,79 @@ static void lt_init_log(void) {
 	uart_print_port = LT_UART_DEFAULT_LOGGER;
 	// default SDK print port
 	serial_init(&m_LogSerial, LT_UART_DEFAULT_PORT, CFG_UART_BAUDRATE_LOG, NULL);
+}
+
+#ifdef FLASH_TUYA_KV_OFFSET
+static bool lt_tuya_kv_read_sta_mac(uint8_t *mac_out) {
+	static const uint8_t IT_MAGIC[8]	= "it_magic";
+	static const uint8_t STA_MAC_KEY[9] = "6_sta_mac";
+
+	const uint32_t kv_base = FLASH_TUYA_KV_OFFSET;
+	const uint32_t kv_size = FLASH_TUYA_KV_LENGTH;
+	uint8_t entry[31];
+	bool found = false;
+	uint32_t off;
+
+	for (off = 0; off + sizeof(entry) <= kv_size; off++) {
+		hal_flash_read(kv_base + off, 8, entry);
+		if (memcmp(entry, IT_MAGIC, 8) != 0)
+			continue;
+		hal_flash_read(kv_base + off + 8, 23, entry + 8);
+		uint16_t val_size = entry[10] | ((uint16_t)entry[11] << 8);
+		if (val_size != 6)
+			goto next;
+		if (memcmp(entry + 16, STA_MAC_KEY, 9) != 0)
+			goto next;
+		if (memcmp(entry + 25, WIFI_MAC_FACTORY_DEFAULT, 6) == 0)
+			goto next;
+		{
+			bool all_ff = true;
+			for (int i = 0; i < 6; i++)
+				if (entry[25 + i] != 0xFF) {
+					all_ff = false;
+					break;
+				}
+			if (all_ff)
+				goto next;
+		}
+		memcpy(mac_out, entry + 25, 6);
+		found = true; // keep scanning -- last entry wins in BLK_V1.0 log
+	next:
+		off += 7; // outer off++ gives total advance of 8
+	}
+	return found;
+}
+#endif // FLASH_TUYA_KV_OFFSET
+
+static void lt_init_unique_mac(void) {
+	uint8_t mac[6];
+	if (SYSPARAM_ERR_NONE != sysparam_sta_mac_get(mac))
+		return;
+	if (memcmp(mac, WIFI_MAC_FACTORY_DEFAULT, 6) != 0)
+		return; // step 1: already unique
+#ifdef FLASH_TUYA_KV_OFFSET
+	if (lt_tuya_kv_read_sta_mac(mac)) {
+		LT_I(
+			"Restored Tuya factory MAC: %02X:%02X:%02X:%02X:%02X:%02X",
+			mac[0],
+			mac[1],
+			mac[2],
+			mac[3],
+			mac[4],
+			mac[5]
+		);
+		sysparam_sta_mac_update(mac);
+		mac[0] |= 0x02; // locally-administered bit
+		sysparam_softap_mac_update(mac);
+		return;
+	}
+#endif // FLASH_TUYA_KV_OFFSET
+	if (ln_generate_random_mac(mac) != 0)
+		return;
+	LT_I("Generated random unique MAC: %02X:%02X:%02X:%02X:%02X:%02X", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
+	sysparam_sta_mac_update(mac);
+	mac[0] |= 0x02;
+	sysparam_softap_mac_update(mac);
 }
 
 void lt_init_family() {
@@ -41,6 +127,8 @@ void lt_init_family() {
 
 	// init system parameter
 	sysparam_integrity_check_all();
+	// randomize MAC if still at factory default (first boot after flash)
+	lt_init_unique_mac();
 
 	ln_pm_sleep_mode_set(ACTIVE);
 	// ln_pm_always_clk_disable_select(CLK_G_I2S | CLK_G_WS2811 | CLK_G_SDIO);


Bl00d-B0b:
**Real-world validation — migration from `WiFi.setMacAddress()` workaround**

Tested on DS-101JL (BSEED) dual gang wall switches with LN882H (ln-cb3s-v1.0).

Previously these devices used a manual `WiFi.setMacAddress()` workaround in ESPHome config to assign unique MACs, since all devices boot with the factory default `00:50:C2:5E:10:88`. LibreTiny's `WiFi.setMacAddress()` implementation persists the MAC to sysparam KV via `sysparam_sta_mac_update()`.

After flashing firmware with this fix applied:
- `lt_init_unique_mac()` reads the sysparam KV MAC
- Finds it is **not** the factory default (already set by the previous workaround)
- Returns immediately — preserving the existing unique MAC, no flash scan needed

Removing the `WiFi.setMacAddress()` workaround from config and reflashing via OTA: devices retain the same MACs across reboots and OTA updates, confirmed via ARP.

This means the fix handles all cases correctly:

1. **Device with existing unique MAC in sysparam KV** (previous workaround, or any boot after the first) — step 1 fast path, no action taken
2. **Fresh device, Tuya KV intact** (standard UART flash or Kickstart/cloudcutter OTA from Tuya stock) — step 2 restores the Tuya factory-assigned MAC from BLK_V1.0 KV; no new address generated
3. **Fresh device, Tuya KV erased** (UART flash with full-chip erase) — step 3 generates a unique MAC via hardware TRNG as fallback

In cases 2 and 3 the resolved MAC is persisted to sysparam KV at `0x1E0000` (outside the OTA partition, never erased by OTA), so every subsequent boot returns in step 1 immediately.

kuba2k2:
See my comments here:
https://github.com/libretiny-eu/libretiny/issues/338#issuecomment-3805949114
https://github.com/libretiny-eu/libretiny/issues/338#issuecomment-3806768732
https://github.com/libretiny-eu/libretiny/issues/338#issuecomment-3806930864
https://github.com/libretiny-eu/libretiny/issues/338#issuecomment-3819613009

TL;DR: Generating a MAC address randomly is a workaround. A real solution would be to adjust the partition layout, or - as mentioned in the linked thread - create a separate board definition for Tuya devices.
Then, LibreTiny should *read* the address from that KV store. No random generation or writing new addresses should be needed, as it's already there on every Tuya device.
It seems to be correct that Tuya firmware generates it randomly and writes it to storage - and we're here to read it from there, instead of generating new ones.

Bl00d-B0b:
### Flash dump analysis — MAC recovery across all use cases

I analyzed 5 unique 2 MB flash dumps from Tuya LN882H devices (plus verified binary details of the OTA write path in LibreTiny itself) to confirm the approach and document why no partition changes are needed.

---

#### Devices analyzed

| Dump | Device | State before dump | Active `6_sta_mac` |
|------|--------|-------------------|--------------------|
| D1 | DS-101JL wall switch | Tuya-provisioned, ~1 month in use | `00:33:7A:2B:86:95` |
| D2 | DS-101JL wall switch | Tuya-provisioned, briefly | `BC:35:1E:BE:41:EB` |
| D3 | DS-101JL wall switch | Never provisioned | `00:33:7A:2B:84:1D` |
| D4 | DS-101JL wall switch | Never provisioned | `00:33:7A:2B:87:C6` |
| GU10 | Tuya GU10 bulb (from #338) | Tuya-provisioned | `C0:F8:53:CF:3A:83` |

---

#### BLK_V1.0 KV structure — confirmed across all devices

Every device has the same `6_sta_mac` layout in its first KV block:

| Offset | Entry | Value | Note |
|--------|-------|-------|------|
| `0x1C51EB` | `6_sta_mac` #1 | `00:50:C2:5E:10:88` | Factory sentinel (`sysparam_factory_setting.h`) |
| `0x1C5423` | `6_sta_mac` #2 | device-unique | Active MAC — last entry wins in BLK_V1.0 log |

Important: BLK_V1.0 keys are not null-terminated. The format is:

```
it_magic[8] | crc16[2] | val_size_u16_le[2] | prev_ptr_u32_le[4] | key_raw[N] | value_raw[val_size]
```

A parser must use known key lengths, not null scanning.

The unique MAC is factory-assigned, not provisioning-assigned. D3 and D4 were never connected to a Tuya account (no `ty_ada_psk` WiFi credential entries) yet both carry a unique `6_sta_mac` at `0x1C5423`. The MAC is written at manufacturing time.

Provisioned devices additionally accumulate `ty_ada_psk` entries (139 B each, home WiFi SSID+PSK in plaintext) appended on each reconnect — D1 has 7 copies after ~1 month. This is normal BLK_V1.0 compaction behaviour and does not affect the MAC entries.

---

#### Why the Tuya KV at 0x1C5000 is preserved — no partition changes needed

The Tuya KV occupies `0x1C5000`–`0x1C9000` (4 x 4 KB blocks), which falls inside LibreTiny's OTA partition (`0x133000`–`0x1DD000`). At first glance this looks like a problem. In practice it isn't, for both flash paths:

**OTA (Kickstart / cloudcutter / LibreTiny OTA update)**

`uf2_write()` uses per-block addressing:

```c
uint32_t offset = part->offset + block->addr;  // physical = OTA_base + UF2_addr
flash->ops.erase(offset, block->len);           // only this sector is erased
```

The current LibreTiny firmware for LN882H contains UF2 blocks with addresses `0x000000`–`0x081700` (~530 KB). To reach `0x1C5000`, a block address of `0x1C5000 - 0x133000 = 0x92000` would be needed. The firmware is 66 KB short of that threshold. No UF2 block targets `0x1C5000`; the sector is never erased.

**UART full-flash (`image_firmware.0x000000.bin`)**

The UART flash image is ~548 KB and covers physical addresses `0x000000`–`0x088000`. The flash tool erases and writes only those sectors. `0x1C5000` is not covered.

**Conclusion:** The Tuya KV is naturally preserved under both flash methods as long as the firmware stays below ~592 KB (~66 KB of headroom remaining). Shrinking the OTA partition to protect this region is not recommended — the trade-off (680 KB → 584 KB, losing 96 KB of firmware capacity) is not worth it for MAC address preservation alone when a software solution covers all realistic flash scenarios.

---

#### Three use cases — MAC resolution behaviour

| Use case | 0x1C5000 on first LT boot | Result |
|----------|--------------------------|--------|
| UART flash (standard, no full-chip erase) | Intact | Step 2: Tuya MAC restored |
| Kickstart / cloudcutter OTA from Tuya | Intact | Step 2: Tuya MAC restored |
| UART flash with full-chip erase | All-FF | Step 3: TRNG random MAC |
| LibreTiny to LibreTiny OTA | Irrelevant | Step 1: MAC already in sysparam KV |

In all cases the MAC is persisted to sysparam KV at `0x1E0000` (outside OTA, never erased), so every subsequent boot hits step 1 immediately.

kuba2k2:
It's a good thing that the MAC address is not erased during OTA updates - yet. If someone builds firmware that's larger than that 0x92000, then it definitely will get erased.

Tuya designed such a partition layout where 0x1C5000 was used for KV. But they also use OTA.

> Shrinking the OTA partition to protect this region is not recommended — the trade-off (losing OTA capacity) is not worth it for MAC address preservation alone

That doesn't quite make sense.
If we shrink OTA, we lose some capacity - not a problem since firmware is usually smaller, right?
If we don't shrink OTA, we *might* lose the MAC address and other calibration parameters if someone builds a large firmware - seems like a problem to me.

Bl00d-B0b:
### Design reflection: approach alignment and partition trade-off

#### On the Tuya KV reading approach

The current PR satisfies the stated preference from #338 and the PR comment above: step 2 reads the factory-assigned MAC directly from the Tuya BLK_V1.0 KV store. The TRNG (step 3) is a fallback only for the edge case where the KV region is erased, which happens exclusively with a full-chip erase UART flash. Under all normal flash paths (standard UART flash, Kickstart, cloudcutter OTA) the Tuya KV is intact and the factory MAC is restored without any random generation.

It is also worth noting that users migrating from a `WiFi.setMacAddress()` workaround — which is the most common real-world path — never reach step 2 at all. Their MAC is already unique in sysparam KV and step 1 returns immediately, preserving whatever address was previously set regardless of whether it matches the Tuya factory MAC. In practice this means the factory MAC recovery in step 2 only matters for a completely clean first flash with no prior MAC set, and even then the end result (a stable unique MAC persisted to sysparam) is identical whether it came from the Tuya KV or from TRNG.

#### On repartitioning

Adjusting the partition layout to protect `0x1C5000` would shrink the OTA partition from **680 KB to 584 KB** — a permanent 96 KB reduction in maximum firmware size paid by all users, in exchange for protecting against a risk that only materialises if firmware ever exceeds ~592 KB. Current firmware is ~530 KB with 66 KB of headroom, and the resolved MAC is persisted to sysparam KV at `0x1E0000` which is outside the OTA partition and safe permanently regardless of firmware size.

There is also a practical deployment problem: the partition table sits at `0x006000`, entirely outside the OTA partition (`0x133000`). OTA updates never touch it. Changing the partition layout therefore requires a full UART reflash of every device — it cannot be rolled out over the air. For devices already installed (e.g. wall switches mounted inside walls), this means physical access to each unit individually. That is not a realistic migration path.

Additionally, even with repartitioning there is no guarantee that future Tuya firmware releases will keep the MAC address at the same offset within the KV region. The BLK_V1.0 layout and the `6_sta_mac` key placement at `0x1C5423` are observed facts from current devices — not a documented stable contract. If Tuya changes their factory provisioning process, the hardcoded offset assumption breaks regardless of partition protection.

#### Recommendation

Given the above, I would suggest following the same KISS approach already used by OpenBeken for LN882H: sysparam fast path → TRNG. Their implementation in [`wifi_init_sta()` (src/hal/ln882h/hal_wifi_ln882h.c, lines 318–330)](https://github.com/openshwprojects/OpenBK7231T_App/blob/main/src/hal/ln882h/hal_wifi_ln882h.c#L318-L330) does exactly this — check sentinel, call `ln_generate_random_mac()`, persist to sysparam. LibreTiny is itself a new approach to these devices and should follow the same established pattern. Keeping the MAC resolution simple and dependency-free is more maintainable long term. Once the MAC is written to sysparam KV on first boot it is stable forever — whether it came from Tuya's factory or from TRNG makes no practical difference for a device behind NAT.

#### Summary

| Option | Factory MAC preserved | OTA partition size |
|--------|----------------------|--------------------|
| This PR (KV read + TRNG fallback) | Yes | 680 KB |
| TRNG only (KISS, recommended) | No — generates new random MAC | 680 KB |
| Repartition | Yes (current Tuya layout only) | 584 KB |

The static IP fix (`WiFiEvents.cpp`, PR #375) is included in this PR and would remain regardless of which MAC approach is chosen — that part is independently useful and has been a long-standing gap for LN882H.

kuba2k2:
I'm currently short on time, so I can't really look deeper into this issue - perhaps I'll be able to in a few days from now.

Bl00d-B0b:
Hi @kuba2k2,

I appreciate the feedback and want to share where I stand, because I don't fully agree with the direction the discussion has taken.

For context: my original implementation used exactly the TRNG approach — check sysparam sentinel, generate via hardware TRNG, persist. I extended it to include Tuya BLK_V1.0 KV reading in response to the review feedback. After going through the full analysis I'm now coming back to the original approach, and here is why.

**On repartitioning** — not a viable option. The partition table lives at `0x006000`, outside the OTA partition, so it can never be updated over the air. Any device already installed inside a wall or ceiling would need physical UART access. That rules it out for a fix that affects every LN882H device.

**On a separate Tuya board definition** — this is the same hardware regardless of original firmware. A board split adds maintenance overhead without solving the root problem.

**TRNG is the right solution.** OpenBeken uses exactly this approach and it works. The MAC is generated once on first boot, persisted to sysparam KV at `0x1E0000` (outside OTA, survives all flash operations), and never changes again. Simple, dependency-free, and proven.

The detailed BLK_V1.0 flash analysis already in this PR can live as **documentation**: we know exactly where the original factory MAC sits (`0x1C5423`, key `6_sta_mac`) and what other parameters (power calibration etc.) are stored nearby. A user who wants to preserve their original Tuya-assigned MAC can take a firmware backup before flashing LibreTiny, read the address from that known offset, and set it once via `WiFi.setMacAddress()` in their ESPHome config. After the first boot it is persisted to sysparam KV and stays there permanently — the manual step is only needed once, and can then be removed from the config entirely. The same approach applies to other device-specific parameters stored in that region.

---

**Wider context — full LN882H platform effort**

This PR is part of a set of changes that together make LN882H a fully functional LibreTiny platform:
- **#375** — static IP support (WiFiEvents.cpp)
- **#381** (this PR) — unique MAC on first boot
- **#382** — BLE 5.1 SDK compiled and linked
- **esphome/esphome#16691** — `ln882h_ble_tracker` component: BLE scanning, BTHome/ATC thermometers, ble_presence, ble_rssi, bluetooth proxy — all tested and working on real hardware (DS-101JL wall switches, LN-CB3S V1.0 modules)

Everything is tested and working from my side. I'm ready to follow the review process properly.

---

**What I need help with**

*Code ownership on ESPHome* — `ln882h_ble_tracker` is a new component and needs a CODEOWNERS entry in ESPHome. Do you know anyone familiar with LN882H or LibreTiny who would be a good fit as code owner? Alternatively, if you are willing to cover it yourself since you already own `libretiny/*` there, that would be great — but I'm equally happy with whoever you think is the right person.

*Tests and reviewers* — ESPHome PR esphome/esphome#16691 has a `needs-tests` label. I need to add `tests/components/ln882h_ble_tracker/` YAML test files (compile-only, same pattern as other LibreTiny components). Do you know any contributors or LN882H users who could help review or validate on real hardware? I can provide ready-to-flash firmware builds.

---

On a broader note — BLE support is a genuinely significant addition to LibreTiny. LN882H has a full BLE 5.1 stack built into its SDK and it works well. Once this lands, the same approach could be extended to the BK7231N/T chipset which also has BLE hardware. Having passive BLE scanning and BTHome sensor support on these cheap widely-deployed chips opens up a lot of home automation use cases that weren't possible before. I think it's worth getting right.

Thanks for the project and for your time.

kuba2k2:
First, "repartitioning" in this case doesn't mean modifying the partition table at 0x6000. I don't know what that table defines, but LibreTiny doesn't use it - instead it has its own, defined in board JSON files.
As is the case for BK72xx, pointing LT at the right calibration data offset should make the SDK use that data. So whatever's defined in the JSON file is used by LT, not what's written at 0x6000. That also allows to adjust the offsets easily in YAML by end users, should devices have vendor-specific or customized firmware.

Second - separate Tuya board definitions are already going to be implemented for BK7238 and perhaps even BK7231N. I know it's "just" the firmware and hardware stays the same (though even that's not always the case). Truth is, big vendors like Tuya use custom partitioning, sometimes custom encryption keys (like on Beken), modified bootloaders (which require specific support), and have their own KV storage in the flash (which should be avoided by LT to keep its original contents). That's all defined by a specific board JSON - and since virtually all boards use base JSON files as well, there's minimal maintenance overhead anyway. And it makes the board list way less ambiguous.
Most smaller vendors, in contrast, won't change their partition layouts, so MAC address will be in OTP or in KV storage somewhere else. Sometimes they might have a real OUI and not just randomly generated bytes, and we want to preserve that. Some people also use OEM devboards to either develop LibreTiny, ESPHome, or their own devices.

An lastly - what OpenBeken does is not considered the golden standard of what we should do. I've seen their code and odd solutions to simple problems way too many times, and many of them just aren't good enough to be used in LibreTiny. Take BK7238 support for example - instead of providing proper OTA support on Tuya bootloaders, they just overwrite it with a OEM bootloader, which leaves devices with mismatched bootloaders and calibration partition layouts, and people migrating from OBK to LT are having issues due to that.

---

I know there are more PRs related to LN882H support. It's a lot to review, and I have limited free time. Also note that I write these comments and review code without using AI agents.

---

As for the code owners, you should probably assign yourself - that's usually what people do. The one who creates code is the owner - seems simple. I definitely wouldn't be a good fit, since I haven't ever flashed a single LN882H device yet.

Bl00d-B0b:
Hi @kuba2k2,

This PR has been updated to implement the board JSON approach you suggested. Here is a summary of what was done.

**Board JSON (your requested architecture)**

- `boards/_base/lightning-ln882hki-tuya.json` — a new base fragment that adds a `tuya_kv` flash section (`0x1C5000+0x4000`). The board generator produces `FLASH_TUYA_KV_OFFSET = 0x1C5000` and `FLASH_TUYA_KV_LENGTH = 0x4000` for any board that includes this base.
- `boards/generic-ln882hki-tuya.json` — a new buildable board for Tuya-manufactured LN882H devices, inheriting the standard `generic-ln882hki` chain plus the new Tuya base.

Non-Tuya boards (`generic-ln882hki`) are not affected.

**`lt_init.c` — board-aware MAC recovery**

The WiFi STA MAC resolution now uses `FLASH_TUYA_KV_OFFSET` from the board JSON as the gate:

1. sysparam KV already unique → fast path, nothing to do
2. `#ifdef FLASH_TUYA_KV_OFFSET` (Tuya boards only): scan BLK_V1.0 KV store for `6_sta_mac` entry → restore factory MAC
3. TRNG fallback — non-Tuya boards, or Tuya KV erased

The factory sentinel `{0x00, 0x50, 0xC2, 0x5E, 0x10, 0x88}` is defined once at file scope (`WIFI_MAC_FACTORY_DEFAULT`) and shared between both functions. The KV offset and region size come entirely from the board JSON — no hardcoded addresses in C.

---

**BLE MAC (related work)**

The companion BLE PR #382 assigns each device a unique BLE MAC the same way — **without any Tuya KV access**. Across all 4 dumps the factory BLE MAC is a fixed function of the WiFi MAC (`BLE = WiFi + 1`), so #382 adds a platform helper `lt_ble_mac_get()` (`cores/lightning-ln882h/base/api/lt_ble.c`) that **derives** it from the WiFi STA MAC instead of reading `24_ble_mac` from the KV.

It takes the WiFi MAC (recovered/persisted by this PR's `lt_init`), increments the low 24-bit NIC by 1 with wraparound (OUI unchanged), MSB-first. No `FLASH_TUYA_KV_OFFSET` — works on any board, including fresh devices with an erased KV where there is no `24_ble_mac` to read. The `ln882h_ble_tracker` component (esphome/esphome#16691) calls the helper, persists the result via the SDK's `ln_kv_ble_addr_store()`, and passes it to `rw_init()`.

Confirmed factory relationship from the dumps: WiFi `…:95 / …:EB / …:1D / …:C6` → BLE `…:96 / …:EC / …:1E / …:C7`.

---

Thanks again for guiding the architecture — the board JSON approach is clearly the right design.

Bl00d-B0b:
Applied the new `lightning-ln882hki-tuya` base to `boards/ln-cb3s-v1.0.json` as well.

The LN-CB3S is the exact Tuya LN882H board this PR''s MAC recovery was validated on (DS-101JL). It was added back in #372 — before the `tuya_kv` base existed — so it never inherited `FLASH_TUYA_KV_OFFSET` and the fix did not actually reach it. This `_base` addition makes MAC recovery a property of the board, matching `generic-ln882hki-tuya`.

No variant regeneration needed: the `tuya_kv` flash region is resolved at build time (`builder/utils/flash.py`); the generated `boards/variants/ln-cb3s-v1.0.h` carries only pin definitions.

kuba2k2:
After digging a little bit deeper into this issue, I found out that no special workarounds are needed at all. As I expected, it all comes down to partition layout differences between SDK/OEM builds and Tuya firmware builds.

The partition at 0x1C5000 is called `kv` - it's **not** Tuya-specific. In fact, the code for reading (and writing) the keys stored there is bundled with the SDK, and already compiled-in with LibreTiny - it's even already used by the Wi-Fi library! Whenever LibreTiny initializes Wi-Fi, it reads the MAC address from that `kv` partition, without any special handling needed.

The whole reason why devices assume the default MAC address is a different partition layout. Take a look:
- [SDK partition layout](https://github.com/libretiny-eu/framework-lightning-ln882h/blob/master/project/wifi_mcu_basic_example/cfg/flash_partition_cfg.json)
- [Tuya partition layout](https://github.com/tuya/TuyaOpen-ln882h/blob/master/ln882h_os/ln882h/cfg/flash_partition_cfg.json)

After just changing the offset to 0x1C5000, the device starts using the right MAC address.

---

I went ahead and committed the partition table refactor to `master` already. It now includes separate `generic` boards for SDK and Tuya firmwares. All other (non-generic) boards now use the Tuya layout.

PR #382, opened 2026-05-28
[lightning-ln882h] Add BLE SDK support

Bl00d-B0b:
## Summary

LN882H is a WiFi+BLE combo SoC — every chip variant has BLE hardware, but the BLE stack was not wired into the LibreTiny build system. This PR adds it using the same `queue.AddLibrary()` pattern used by the rest of `lightning-ln882h.py`, and gates compilation on a `CFG_SUPPORT_BLE` flag so WiFi-only builds stay lean.

### `cores/lightning-ln882h/base/config/proj_config.h`
- Adds `#define CFG_SUPPORT_BLE 0` (default off; enable via `custom_options.proj_config: "CFG_SUPPORT_BLE=1"`).

### `builder/utils/config.py` — `env_load_config` parser fix
- `split(None, 2)` → `split(None, 1)` + skip bare defines (include guards, macros with no value). The old code crashed on `#define _PROJ_CONFIG_H_` and on multi-word values like `((size_t)(160 * 1024))`.

### `builder/utils/env.py` — `proj_config` custom_options alias
- Adds `"proj_config": "proj_config.h"` to the headers dict in `env_parse_custom_options`, so builds can enable BLE via `custom_options.proj_config: "CFG_SUPPORT_BLE=1"`.

### `builder/family/lightning-ln882h.py`
- Calls `env.LoadConfig("$FAMILY_DIR/base/config/proj_config.h")` (same pattern as beken-72xx.py).
- Merges any `custom_options.proj_config` user overrides into `env["CONFIG"]` so `env.Cfg("CFG_SUPPORT_BLE")` reflects the override at build time.
- New `ln882h_ble` library block, active when `env.Cfg("CFG_SUPPORT_BLE")`:
  - **Compiled sources** from `$SDK_DIR/components/ble/`: arch init, port glue, profiles common, GAP (scan/misc/advertising), GATT common, connection/device manager, SMP, store, event, import — 16 C files total.
  - **Include paths**: adds `+<.>` (base_dir itself, required so sources using `#include "ble_arch/arch.h"` resolve) plus all BLE SDK subdirectory paths.
  - **Default `ble_app_user_cfg.h`** at `cores/lightning-ln882h/ble/`, exposed via `AppendPublic(CPPPATH)`. Applications that prepend their own include path shadow this default automatically.
  - **Pre-compiled BLE stack** `libln882h_ble_full_stack.a` linked via `-Wl,--whole-archive`.
  - **`CFG_SUPPORT_BLE=1`** added to public CPPDEFINES.

### `cores/lightning-ln882h/base/api/lt_ble.{c,h}` — unique BLE MAC

The BLE SDK has no factory BLE address of its own — `ln_kv_ble_pub_addr_get()` returns the fixed placeholder `BLE_DEFAULT_PUBLIC_ADDR` (`00:50:C2:00:00:00`) on a blank device, identical on every chip. New platform helper `lt_ble_mac_get(uint8_t out[6])` derives a unique address from the WiFi STA MAC: low 24-bit NIC + 1 (wraparound, OUI unchanged), MSB-first. This reproduces the `BLE = WiFi + 1` pairing confirmed across the Tuya LN882H flash dumps (libretiny-eu/libretiny#381) **without** reading the Tuya KV, so it works on any board. Consumers (e.g. esphome/esphome#16691) call it, persist via the SDK's `ln_kv_ble_addr_store()` (`"2_ble_addr"`), and pass it to `rw_init()`.

## Flash / RAM impact

`libln882h_ble_full_stack.a` is compiled with `-ffunction-sections`. Because `--gc-sections` is always active, functions not reachable from the entry point are stripped. The overhead depends on what the application calls — measured on real hardware (DS-101JL / generic-ln882hki, ESPHome 2026.5.3, LibreTiny 1.12.1), identical config differing only by the BLE layer:

| Build | Flash | RAM |
|---|---|---|
| WiFi-only (`CFG_SUPPORT_BLE=0`) | 530,672 B (43.2%) | 97,000 B (32.1%) |
| BLE SDK, no app code (`CFG_SUPPORT_BLE=1`, no tracker) | 535,616 B (43.6%) | 97,440 B (32.3%) |
| **Delta (startup hooks only)** | **+4,944 B (+4.8 KB)** | **+440 B** |

A build that actively uses the BLE stack will retain all scanner/GAP/GATT code it calls — the savings from disabling BLE in that case will be correspondingly larger.

## Related

esphome/esphome#16691 adds LN882H BLE scanner support and depends on this PR.


Bl00d-B0b:
Gentle ping on this one 🙂 — it's been open a few weeks and is the dependency blocking BLE on LN882H.

**What it does:** adds the BLE SDK to the `lightning-ln882h` build — registers the `components/ble` sources and links the already-shipped `libln882h_ble_full_stack.a`, following the same pattern `beken-72xx.py` already uses for BK.

**Why it's safe to merge:**
- Fully **opt-in** — gated on `CFG_SUPPORT_BLE`, which defaults to **`0`** in `proj_config.h`. Existing WiFi-only builds are completely unaffected.
- With the flag off, the stack is built with `-ffunction-sections` + `--gc-sections`, so a WiFi-only build grows by only ~5 KB (startup hooks).
- CI is green (clang-format + python lint), mergeable with no conflicts, and self-contained — no changes to other families.

It's the gating dependency for the ESPHome LN882H BLE proxy/tracker (esphome/esphome#16691), so merging would unblock that downstream work. No rush at all, but a review whenever you have a moment would be hugely appreciated 🙏 Thanks for all your work on LibreTiny!

kuba2k2:
1. The change in `config.py` broke compilation on BK72xx (`KeyError: None` in `SOC_BDK[SOC]`).
2. Instead of merging custom options in `lightning-ln882h.py`, I modified `env.Cfg()` to read from user options if a header name was passed.

Note: WIP comment, further review will follow.

kuba2k2:
Moved `ble_app_user_cfg.h` to `config/`. Applications can either prepend a higher-priority include path that overrides `config/`, or use something like:

```yaml
esphome:
  platformio_options:
    custom_options.ble_app_user_cfg#h:
      BLE_DEFAULT_ROLE=BLE_ROLE_PERIPHERAL
```
(untested, I don't know the exact syntax for YAML, testing appreciated)

kuba2k2:
Updated the comment in `proj_config.h`, since the #define is not actually used by the SDK (it's only checked by LibreTiny's builder script).

kuba2k2:
Removed `lt_ble.c`/`.h`, since LibreTiny has no BLE API of its own.

If the SDK API is to be used, then leave it that way.

Note that no regeneration of BLE MAC address is needed. The source code comment:

> The BLE SDK has no factory BLE address of its own — ln_kv_ble_pub_addr_get()
> returns a fixed placeholder (BLE_DEFAULT_PUBLIC_ADDR, 00:50:C2:00:00:00) on a
> blank device, identical on every chip. [...]

is not true. The function `ln_kv_ble_app_init()` calls `ln_kv_ble_addr_load()`, which reads the BLE address using `btparam_load()`, which in turn calls `ln_kv_get()` with the key name `"2_ble_addr"`. Since the partition layout of KV storage was corrected in 9c5111248f6582b1def26d7244dfe685a9d906b4, devices can access the storage properly, and many of them already have a correct, persistent MAC address stored.

In case the address isn't stored, then yes, the SDK will use a default one. ESPHome (or whatever is using the SDK library **directly**) should handle this - because, again, LibreTiny has no BLE API.

kuba2k2 (approved):

PR #386, opened 2026-06-02
[boards] Fix WB02A pin 4 mapping

### LINKED ISSUE #384: WB02A pinout maps PA05 twice; pin 4 should be PA06
The WB02A board definition currently maps PA05 twice.

Once [here](https://github.com/libretiny-eu/libretiny/blob/af7e7e82c66e150a17b75a446ba034b1f204b156/boards/_base/pcb/wb02a.json#L54-L57). I checked on my hardware and found that module pin 4 is the incorrect one, instead of:
https://github.com/libretiny-eu/libretiny/blob/af7e7e82c66e150a17b75a446ba034b1f204b156/boards/_base/pcb/wb02a.json#L21-L24

It should be PA06:
   ```json
   "4": {
     "IC": 14,
     "ARD": "D10"
   },
```

On the original PR #332 it was mentioned that a dump might be useful, here's the [one I got](https://github.com/user-attachments/files/28467095/ln882h_dump.bin.tar.gz). The hardware is a different model of smart plug ([Aliexpress link](https://aliexpress.com/item/1005005481310738.html)).

Happy to open a PR if that would help.

Thanks for the project!

enosh:
Closes #384.

This fixes WB02A module pin 4, which was previously mapped to IC pin 13 / PA05, duplicating pin 11. But is connected to IC pin 14 / PA06.

The diff looks so big because the extra digit in 10 forces realignment.

Here's what the new pinout look like:
wb02a


kuba2k2:
Thanks. It seems like on some of these boards (for example: https://www.elektroda.com/rtvforum/topic4169195.html) the silkscreen print saying `A6` looks almost like `A5`, that's probably why #332 had the mistake to begin with.

Issue #393, opened 2026-07-14
BK7238: GPIO edge interrupts never fire (pulse_counter / hlw8012 / BL0937 read 0)


blafois:
### Summary

On the freshly-added **BK7238** target, **GPIO edge interrupts never fire**. `attachInterrupt()`-based consumers (ESPHome `pulse_counter`, `hlw8012`/BL0937 energy metering) receive **zero edges**, even with a strong, continuous signal on the pin. The same wiring/config works on BK7231N.

This makes any interrupt/pulse-driven peripheral (energy meters, pulse counters, etc.) non-functional on BK7238, while polling-based GPIO (button input, relay/LED output) works fine.

### Environment

- LibreTiny **v1.13.0** (latest release; `master` is identical) — BK7238 support from #360, BDK 3.0.78
- ESPHome **2026.8.0-dev** (dev branch, native BK7238 support)
- Board: `generic-bk7238-tuya`
- Device: LSC Smart Plug w/ energy meter (Tuya T1-2S-NL module, BK7238, BL0937 metering chip)
- Flashed via UART (`ltchiptool`); app confirmed booting (correct build timestamp in logs)

### What works vs. what doesn't

| Function | GPIO | Mechanism | Works? |
|---|---|---|---|
| Relay | P24 | digital output | ✅ |
| Status LEDs | P9, P11 | digital output | ✅ |
| Button | P26 | **polling** input | ✅ |
| BL0937 power meter | CF=P6, CF1=P8, SEL=P1 | **interrupt** (pulse) | ❌ reads 0 |

### Minimal reproduction

Bare `pulse_counter` on the BL0937 CF (power) and CF1 (voltage/current) pins:

```yaml
sensor:
  - platform: pulse_counter
    name: "CF pulses (P6 power)"
    pin: P6
    update_interval: 5s
  - platform: pulse_counter
    name: "CF1 pulses (P8 V/I)"
    pin: P8
    update_interval: 5s
```

Plug energized on **live mains, relay ON, ~250 W resistive/inductive load (a running drill) drawing current**. A BL0937 CF output at 250 W should pulse at several hundred Hz.

**Observed** — flat zero for 25+ seconds on both pins:

```
[pulse_counter:200]: 'CF pulses (P6 power)':  Retrieved counter: 0.00 pulses/min
[pulse_counter:200]: 'CF1 pulses (P8 V/I)':   Retrieved counter: 0.00 pulses/min
(repeats every 5 s, always 0.00)
```

[...]

PR #394, opened 2026-07-15
[beken-72xx] Fix digitalRead() returning a stale level inside GPIO ISR (pulse_counter/hlw8012 read 0)

blafois:
## Problem

On **BK7238**, calling `digitalRead()` **inside a GPIO interrupt handler** can return a stale level. For a fast pulse train the edge interrupt reaches the ISR after the pin has already returned to its idle level (interrupt latency exceeds the pulse high-time), so the register read no longer reflects the edge that fired. I have only observed and verified this on BK7238 — BK7231x with BL0937 is reported to work fine, so the fix is now scoped to BK7238 only (see **Scope** below).

This breaks any consumer that determines the edge direction by reading the pin inside the ISR. The most common case is **ESPHome's `pulse_counter`** (and everything built on it — `hlw8012`, `cse7766`, …), whose ISR does:

```cpp
PulseCounterCountMode mode = arg->isr_pin.digital_read() ? arg->rising_edge_mode : arg->falling_edge_mode;
```

With the default `rising_edge_mode = INCREMENT, falling_edge_mode = DISABLE`, and `digital_read()` always returning LOW inside the ISR, **every edge is treated as falling and discarded → the counter never increments.** On a BK7238 smart plug with a BL0937 energy meter, power/voltage/current all read `0`.

## Root cause (verified on hardware)

I instrumented this extensively on a BK7238 (LSC/Action plug, BL0937) using an always-present ~4 kHz signal on a GPIO:

- Polling `digitalRead()` in a loop: works (reads the toggling level).
- Manual GPIO INT registers + poll `INTSTA`: the edge latches correctly.
- `attachInterrupt()` RISING / FALLING / CHANGE: the ISR **fires correctly** (~2 k/s, ~2 k/s, ~4 k/s).
- **But `digitalRead()` *inside* the ISR returned LOW on 72124/72124 calls** — never the post-edge level.

So the interrupt path is fine; only the in-ISR pin read is unreliable. There is no separate fast GPIO input-data register on this SoC (unlike ESP8266's `GPI`, whose `ISRInternalGPIOPin::digital_read()` reads the live pad register directly), so reading "faster" isn't possible here.

## Fix

The ISR already knows which edge fired: `data->irqMode` holds the armed edge (before it is flipped for CHANGE emulation), which implies the pin's logical level. Cache that level in file-scope state and return it from `digitalRead()` while the pin's handler is executing. No behavior change outside an ISR.

GPIO interrupts share a single IRQ vector with non-reentrant dispatch (`gpio_isr`), so the two file-scope variables cannot race across nested handlers.

## Scope

The workaround is gated behind `#if LT_BK7238` (the family macro auto-defined for `beken-7238` builds). It covers the `lt_isr_level`/`lt_isr_gpio` cache definitions, the edge-cache writes in `irqHandler`, and the short-circuit in `digitalRead()`. On **BK7231x** (and every other BK72xx variant) the macro is undefined, so the entire workaround compiles out and behaviour there is byte-for-byte identical to before — nothing that currently works can regress. If the same in-ISR latency turns up on another chip, the guard can simply be widened.

Verified: a fresh BK7238 ESPHome build (LibreTiny 1.13.0, `generic-bk7238-tuya`) compiles clean with the guards and the metering works on hardware.

## Before / after (BK7238 + BL0937, same pin/signal)

| | before | after |
|---|---|---|
| ESPHome `pulse_counter` (default mode) | **0** pulses/min | **120 000** pulses/min |
| ESPHome `hlw8012` power | **0 W** | **~3065 W** (live, uncalibrated) |
| `hlw8012` current / voltage | 0 A / 0 V | **~9.9 A / ~373 V** (live) |

## Notes

- Refs #393 (full investigation log there).


blafois:
@kuba2k2 ?

kuba2k2:
This is an interesting workaround. However, since we haven't had issues on BK7231x using BL0937, I'd prefer if your changes were scoped to BK7238 specifically, and perhaps could be adapted to other chips later on - we don't want to break something that's working.

A simple #if LT_BK7238 should do the trick.

blafois:
Good call @kuba2k2 — thanks. I've scoped the workaround to BK7238 only via `#if LT_BK7238` guards around the ISR-level cache (`lt_isr_level`/`lt_isr_gpio` definitions, the writes in `irqHandler`, and the read short-circuit in `digitalRead`). On BK7231x the macro is undefined so all of it compiles out and behaviour there is byte-for-byte unchanged; only BK7238 builds pick up the fix. Left a comment noting it's BK7238-specific and can be extended to other chips later if the same latency issue turns up. Force-pushed to the branch.

lorf:
@kuba2k2 There are issues with BK7231x using BL0937, check https://github.com/esphome/issues/issues/5812. I have a similar problem and tested this workaround on my board (BK7231N + BL0973), replacing "#if LT_BK7238" with "#if 1". The results are:

| | Power | Voltage | Current |
| :--- | :--- | :--- | :--- |
| Before, pulse_counter, rising edge INCREMENT, falling edge DISABLE, pulses/min | 0 | 0 | 0 |
| Before, pulse_counter, rising edge DISABLE, falling edge INCREMENT, pulses/min | 1077 | 241856 | 958 |
| Before, non-inverted CF/CF1 | 0W | 0V | 0A |
| Before, inverted CF/CF1 | 29W | 602V | 0.211A |
| After, pulse_counter, rising edge INCREMENT, falling edge DISABLE, pulses/min | 538 | 121200 | 480 |
| After, pulse_counter, rising edge DISABLE, falling edge INCREMENT, pulses/min | 540 | 121680 | 479 |
| After, non-inverted CF/CF1 | 14W | 300V | 0.100A |
| After, inverted CF/CF1 | 14W | 300V | 0.100A |
| Expected | 10.5W | 230V | 0.05A |

So this workaround definitely fixes skipping short BL0937 pulses (or counting them twice if pin is inverted) on BK7231N too. The results are still off for my setup, but it's irrelevant to this PR.

PR #395, opened 2026-07-16
Replace rotten BK72xx Discord links with archive.org

### LINKED ISSUE #320: Linked documentation on the Beken 72xx page is unavailable
The linked documentation on the [Beken 72xx page](https://docs.libretiny.eu/docs/platform/beken-72xx/) ([source](https://github.com/libretiny-eu/libretiny/tree/9114bc4c4961ab38ea41a1fdf21ec79eb97d9254/docs/platform/beken-72xx)) which is hosted on discord is no longer available.
Specifically, the first 4 links on the page.

This is due to a discord change that happened sometime back which makes it so links to uploaded media only last for a limited time.

Those pdfs should be re-uploaded elsewhere, so that they're made accessible.

okh-mzny:
Discord doesn't store files permanently, and so the datasheet links  on the Beken 72xx page broke due to link rot.

These were the only links to cdn.discordapp.com I could find in the repo.

Fixes #320


Cossid:
archive.org is proving to be unreliable as well, as they are currently offline for example.

We'll probably have to see about making a repository for documentation hosting or something instead I think.

PR #399, opened 2026-07-27
[common] Remove __DATE__/__TIME__ from banner macro

bdraco:
Follow up to esphome/esphome#17726, which enabled ccache for ESPHome's LibreTiny builds; the hit rate there is capped by this banner macro, so the remaining fix belongs here.

libretiny.h is included by every source file, and having __DATE__ __TIME__ in LT_BANNER_STR defeats compiler caches (ccache direct mode) and reproducible builds for the whole tree. This moves the compile timestamp out of the banner; lt_main.c logs it right after LT_BANNER(), using LT_LOG directly like the banner does so it prints regardless of LT_LOGLEVEL, and only that one file rebuilds when the time changes.

The intended visible change is that the startup banner no longer contains the compile timestamp; it is printed on its own log line right after the banner. The timestamp stays reachable through a new lt_get_build_timestamp() accessor, which ESP.getFullVersion() appends at runtime, so its result keeps the timestamp without pulling __DATE__ into every TU. The value can be overridden by defining LT_BUILD_TIMESTAMP, letting reproducible builds wire in SOURCE_DATE_EPOCH.

Verified with ESPHome pointing the framework source at this branch, on bk72xx (cb3s) and rtl87xx (bw15); the firmwares contain the new banner and the separate timestamp line, a lambda calling ESP.getFullVersion() compiles and links against the new accessor, and a full rebuild after esphome clean goes from roughly 80 percent ccache hits to 421 of 422, finishing in 18 seconds instead of 72.

The vendor SDKs were also checked; framework-beken-bdk and framework-realtek-amb1 are clean, framework-realtek-ambz2 has three .c files using the macros with single file impact each, and framework-lightning-ln882h defines LN882H_SDK_BUILD_DATE_TIME in mcu/ln882h/ln882h.h, which poisons every LN882x file the same way and needs a one line fix in that repo.

### DIFF
--- a/cores/common/arduino/libraries/inline/ESP/ESP.h
+++ b/cores/common/arduino/libraries/inline/ESP/ESP.h
@@ -50,8 +50,10 @@ class EspClass {
 	/** @copydoc LT_VERSION_STR() */
 	inline String getCoreVersion() { return LT_VERSION_STR; }
 
-	/** @copydoc LT_BANNER_STR() */
-	inline String getFullVersion() { return LT_BANNER_STR; }
+	/** @brief Full version string; the compile timestamp is appended at
+	 * runtime so that __DATE__/__TIME__ stay out of this header, which
+	 * would otherwise defeat compiler caches (see libretiny.h). */
+	inline String getFullVersion() { return String(LT_BANNER_STR ", compiled at ") + lt_get_build_timestamp(); }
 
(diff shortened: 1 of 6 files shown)


esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #400, opened 2026-07-27
[core] Write custom option headers to an env independent include dir

bdraco:
Follow up to esphome/esphome#17726, which enabled ccache for ESPHome's LibreTiny builds; sharing the cache between devices is blocked by this generated header path, so the remaining fix belongs here.

The generated custom option headers (lwipopts.h, sys_config.h and friends) are currently written to ${BUILD_DIR}/include, a path that embeds the PlatformIO env name. That -I flag makes every compile command unique per env, so compiler caches like ccache can never share results between projects that differ only by env name; ESPHome creates one env per device, so nothing is shared between devices.

This writes them to ${PROJECT_BUILD_DIR}/include/<hash> instead, where the hash is derived from the option content. The path no longer contains the env name, so with ccache base_dir the compile commands hash identically across devices with the same options; envs with different options get different directories, so they can never pick up each other's generated headers, and concurrent builds of envs with identical options write identical bytes. The header writes also moved to a with block so a failed build cannot leave a truncated header behind.

Verified with ESPHome pointing the framework source at this branch; a clean bk72xx build succeeds, the headers land in the content keyed directory, 292 dependency files reference them at the new path with none referencing the old one, and a second compile reproduces the same hash with nothing rebuilt.

### DIFF
--- a/builder/utils/env.py
+++ b/builder/utils/env.py
@@ -2,6 +2,7 @@
 
 import json
 import sys
+from hashlib import sha1
 from os import makedirs
 from os.path import isdir, join
 from subprocess import PIPE, Popen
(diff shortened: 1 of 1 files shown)


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Shared include dir leaks generated headers between envs of the same project

esphbot (approved):
> [!TIP]

esphbot (approved):
> [!TIP]

PR #401, opened 2026-08-07
realtek-amb: fix double free and truncated copy of Arduino WiFi event data

Bl00d-B0b:
## Summary

Three defects in the Arduino WiFi event path shared by all `realtek-amb` families (AmebaZ, AmebaZ2, and the AmebaD work in progress). Found while bringing up RTL8720DN: the device hard-faulted inside `memcpy` called from `wifi_indication()`, with `0xdeadbeef` in R0-R3 (heap poison).

### 1. Double free of the event buffer

`WiFiSTA.cpp:157` queues a real `EventInfo *` with `flags == -2`:

```cpp
wifi_indication(WIFI_EVENT_CONNECT, (char *)eventInfo, ARDUINO_EVENT_WIFI_STA_GOT_IP, -2);
```

`wifi_indication()` copies it into `ev->buf`. Then `handleRtwEvent()` frees that copy in the `flags == -2` branch, and `wifiEventTask()` frees the same pointer again immediately afterwards. `wifiEventTask()` now owns and frees the queued copy for every `flags` value; `handleRtwEvent()` frees nothing — one deletion retires the double free on both the queued and the non-queued path.

### 2. Copy length taken from an event id

On the `flags == -2` path `buf_len` carries the Arduino event id, not a byte count, so the `EventInfo` was copied at whatever length that id happened to be. It now copies `sizeof(EventInfo)`.

### 3. Null event info dereferenced

Events raised with no payload (`ARDUINO_EVENT_WIFI_STA_START` and friends in `WiFiGeneric.cpp`/`WiFiAP.cpp`) pass `buf == NULL`, which reached `*eventInfo`. Now handled explicitly.

The copy is also guarded against a `buf` that is not a pointer: the AmebaD driver calls `wifi_indication()` with `buf = 0x1, buf_len = 2` while switching modes, which is what produced the original fault.

## Testing

- Built for `bw15` (AmebaZ2) with this change: SUCCESS, RAM 4.8%, flash 38.5%.
- On RTL8720DN hardware the fault is gone and the device completes association and DHCP; the same code path now runs to `GOT_IP` without heap corruption.

### DIFF
--- a/cores/realtek-amb/arduino/libraries/WiFi/WiFiEvents.cpp
+++ b/cores/realtek-amb/arduino/libraries/WiFi/WiFiEvents.cpp
@@ -6,6 +6,10 @@
 
 #define WIFI_EVENT_MAX_ROW 3
 
+// Lowest address treated as a real pointer in wifi_indication(). Drivers on
+// some families pass a small integer in buf instead of a pointer.
+#define WIFI_EVENT_BUF_MIN_ADDR 0x1000
+
 static xQueueHandle wifiEventQueueHandle = NULL;
 static xTaskHandle wifiEventTaskHandle	 = NULL;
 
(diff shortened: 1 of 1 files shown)


bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Signed guard weakened: negative buf_len now passes the length check
> - When the copy is skipped, buf_len is still forwarded — NULL data with a non-zero length

bdraco:
Sorry, wrong window, didn't mean to drop a review from the bot on this one.

Bl00d-B0b:
Both blockers plus the three non-blocking notes, in e3326f7:

1. **Signed guard** — the sign check now runs before the cast: `(flags == -2 || buf_len > 0)` gates entry, and `(size_t)buf_len` only executes on the positive path. A negative length can no longer widen to `SIZE_MAX`.
2. **Forwarded length** — when no copy survives, the length is zeroed with the pointer (`if (!bufCopy) buf_len = 0;`), so `handleRtwEvent()` and SDK handlers never see NULL data with a non-zero length. The `WIFI_EVENT_STA_DISASSOC` deref path is closed.
3. **Ownership split** — took the simpler model you suggested: the queue owns its copy for every flags value and frees it in `wifiEventTask()`; `handleRtwEvent()` never frees what it is passed. This also removes the pre-existing double free on the non-queued fallback path (`WiFiSTA.cpp` frees its own buffer after the call, as it always did).
4. `0x1000` is now `WIFI_EVENT_BUF_MIN_ADDR` with a comment.
5. Both `malloc` results are checked; a failed event allocation drops the event instead of faulting.

Hardware state: an RTL8720DN (BW16) is running this exact code — boots, associates, DHCP, and serves the ESPHome native API with no heap corruption across repeated reconnects. clang-format 19 clean.

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Zeroing buf_len destroys the Arduino event id on the flags == -2 path

Bl00d-B0b:
All three, in 18eb65d:

1. The zeroing is gated on `flags != -2`, with a comment stating why the id must survive. STA_START/STA_STOP/AP_START/AP_STOP deliver their real ids again — verified by reading the queue consumer: `handleRtwEvent()` takes the `-2` early-return before anything reads `len` as a length.
2. A failed payload allocation now frees the event and returns, so `!bufCopy` means exactly "no payload was supplied".
3. `WIFI_EVENT_STA_DISASSOC` validates `data && len >= 6` before the MAC copy.

Branch is 0 behind master. Compiled for the AmebaD family (shared file); the previous head runs on BW16 hardware — this delta changes only which id the four no-payload events carry.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
Hardware confirmation of the final revision on RTL8720DN: with event-flow instrumentation in the queue path, a no-payload Arduino event now arrives as
```
ind ev=0 len=2 fl=-2  ->  arduino id=2
```
— STA_START (2) delivered with its id intact through the NULL-buf path, the exact case the last review round fixed. Association, DHCP and the ESPHome native API all run on this code.

Bl00d-B0b:
Both notes addressed: the description's ownership bullet now states the final design (task owns and frees the queued copy for every flags value; handleRtwEvent frees nothing), and 0a14b62 adds a log line on both heap-pressure drop paths in wifi_indication() — the event-record malloc and the payload malloc — so a dropped disconnect under memory pressure is visible in the field. The third path (address below WIFI_EVENT_BUF_MIN_ADDR) stays silent deliberately: that is the documented value-in-buf convention (AmebaD passes buf=0x1 on every no-payload Arduino event), so a log there would fire on routine mode changes, not on losses.

esphbot (approved):
> [!TIP]

PR #405, opened 2026-08-21
[realtek-amb] pinRemoveMode(): null the driver pointers after free

### LINKED ISSUE #404: [realtek-amb] pinRemoveMode() use-after-free: pwm/gpio/irq pointers freed but not nulled
**Family:** realtek-amb (observed on RTL8720CF / AmebaZ2, LibreTiny v1.13.0)

`pinRemoveMode()` in `cores/realtek-amb/arduino/src/wiring.c` frees the per-pin driver objects but leaves the pointers dangling:

```c
if ((mask & PIN_PWM) && (pin->enabled & PIN_PWM)) {
    pwmout_free(data->pwm);
    free(data->pwm);          // data->pwm still points at freed memory
    pinDisable(pin, PIN_PWM);
}
```

Same pattern for `data->gpio` and `data->irq`. The attach paths decide whether to (re)initialize by checking the pointer for NULL, not the enabled mask — e.g. `analogWrite()`:

```c
pwmout_t *pwm = data->pwm;
if (!pwm) {                   // stale non-NULL pointer skips re-init
    data->pwm = pwm = malloc(sizeof(pwmout_t));
    pwmout_init(pwm, pin->gpio);
    ...
}
pwmout_write(pwm, percent);   // writes through freed memory
```

So any GPIO↔PWM mode round-trip on a pin is a use-after-free. Repro on RTL8720CF: `analogWrite(pin, x)` → `pinMode(pin, OUTPUT)` → `analogWrite(pin, y)`. First observable symptom is garbage pin state (`[GPIO Err]hal_gpio_pull_ctrl: Invalid pin, port=7 pin=0`), followed by heap corruption and an imprecise bus fault with `0xbeefdead` in registers a few calls later.

This is reachable from ordinary Arduino/ESPHome code: ESPHome's `libretiny_pwm` output walks exactly this path in `update_frequency()` (it drops to `pinMode(INPUT)` and back to force a new PWM period), and ESPHome's IR transmitter did too until we bypassed the wiring layer (esphome/esphome#18579).

Suggested fix: NULL the pointers after `free()` in `pinRemoveMode()` (three lines), or make the attach paths key off `pinEnabled()` instead of pointer-NULL. The lightning-ln882h core's `pinRemoveMode()` has the same free-without-null pattern for `data->gpio` and may deserve the same audit.

Bl00d-B0b:
Fixes #404.

`pinRemoveMode()` frees the per-pin pwm/gpio/irq driver objects but leaves the pointers dangling, and every attach path decides whether to (re)initialize by checking the pointer for NULL rather than the enabled mask — so the next attach after a remove reuses the freed object.

Three-line fix: null each pointer right after its `free()`.

Verified by compile for `bw15` (AmebaZ2) and `bw16` (AmebaD, via the family PR's bench setup); hardware regression coverage on RTL8720DN follows in the ongoing pin bring-up there.

### DIFF
--- a/cores/realtek-amb/arduino/src/wiring.c
+++ b/cores/realtek-amb/arduino/src/wiring.c
@@ -56,17 +56,23 @@ void pinRemoveMode(PinInfo *pin, uint32_t mask) {
 	if ((mask & PIN_GPIO) && (pin->enabled & PIN_GPIO)) {
 		gpio_deinit(data->gpio);
 		free(data->gpio);
+		// attach paths (re)initialize on a NULL pointer, not the enabled
+		// mask - a dangling pointer here means the next attach reuses the
+		// freed object instead of allocating a fresh one
+		data->gpio = NULL;
 		pinDisable(pin, PIN_GPIO);
 	}
 	if ((mask & PIN_IRQ) && (pin->enabled & PIN_IRQ)) {
 		data->irqHandler = NULL;
 		gpio_irq_free(data->irq);
 		free(data->irq);
+		data->irq = NULL;
 		pinDisable(pin, PIN_IRQ);
 	}
 	if ((mask & PIN_PWM) && (pin->enabled & PIN_PWM)) {
 		pwmout_free(data->pwm);
 		free(data->pwm);
+		data->pwm = NULL;
 		pinDisable(pin, PIN_PWM);
 	}
 }


Bl00d-B0b:
Hardware confirmation on RTL8720DN: with this patch riding, a full pin-verification round (repeated pinMode/analogWrite attach-remove cycles across every header GPIO, driven via Home Assistant) ran with no stale-pointer faults; all pins behave correctly through mode transitions.

PR #408, opened 2026-08-22
[beken-72xx] Add BLE bonding partition to prevent bootloader erase

### LINKED ISSUE #407: [beken-72xx] Missing BLE bonding partition bricks device (BDK erases bootloader sector 0)
## Summary

The beken-72xx partition table omits `BK_PARTITION_BLE_BONDING_FLASH`, which the Beken BDK BLE stack requires. On builds that initialise the BLE bond store this erases flash **sector 0 (the bootloader)** and permanently bricks the device. Full downstream analysis and reproduction: esphome/esphome#18646 (BK7238 / Tuya T1-3S, LibreTiny 1.13.0, beken-bdk 3.0.78).

## Root cause

`cores/beken-72xx/base/wraps/BkDriverFlash.c` provides the partition table the BDK reads via `bk_flash_get_info()`:

```c
static const bk_logic_partition_t bk7231_partitions[BK_PARTITION_MAX] = {
    [BK_PARTITION_BOOTLOADER]  = { .partition_start_addr = FLASH_BOOTLOADER_OFFSET,  ... },
    [BK_PARTITION_APPLICATION] = { .partition_start_addr = FLASH_APP_OFFSET,         ... },
    [BK_PARTITION_OTA]         = { .partition_start_addr = FLASH_DOWNLOAD_OFFSET,     ... },
    [BK_PARTITION_RF_FIRMWARE] = { .partition_start_addr = FLASH_CALIBRATION_OFFSET, ... },
    [BK_PARTITION_NET_PARAM]   = { .partition_start_addr = 0x1FF000,                 ... },
    // no [BK_PARTITION_BLE_BONDING_FLASH]
};
```

`BK_PARTITION_BLE_BONDING_FLASH` is in enum range, so `bk_flash_get_info()` returns a pointer to the **zero-initialised** slot (`partition_start_addr == 0`). The BDK bond store (`driver/ble/ble_5_x/ble_pub/app/src/app_sec.c`) then runs, on first boot with blank bond flash:

```c
app_sec_env.flash_bond_ptr = bk_flash_get_info(BK_PARTITION_BLE_BONDING_FLASH);   // -> start_addr == 0
...
flash_ctrl(CMD_FLASH_ERASE_SECTOR, &app_sec_env.flash_bond_ptr->partition_start_addr);   // erases 0x000000
```

→ sector 0 (bootloader) erased → unbootable, also after re-flashing a normal image (which does not write sector 0) and after OTA. The table is shared across all beken-72xx chips, so the zero address is latent everywhere; it surfaces on BK7238 because its BLE **5.2** stack reaches the bond-store erase during `ble_entry()` (the BK7231N 5.1 scan-only path does not, which is why BK7231N BLE proxies have not bricked).

## Proposed fix

1. Add a `[BK_PARTITION_BLE_BONDING_FLASH]` entry to `bk7231_partitions[]`, driven by a per-layout `FLASH_BLE_BONDING_OFFSET` macro, so `partition_start_addr` is never 0.
2. Reserve a dedicated 4 KB `ble_bonding` sector in the board flash layouts that BLE builds use. The full-flash Tuya layout (`boards/_base/beken-7238-tuya.json`) currently allocates all 2 MB — `userdata 0x1EC000+0x14000`, `tuya 0x1F5000+0xB000`, `net_param 0x1FF000` — leaving **no free sector**, so even the BDK's own 2M default (`0x1f0000`) overlaps `userdata`. A sector needs to be carved out and kept clear of app/OTA/userdata for these layouts.

I don't have BK7238 hardware and a wrong partition change could itself brick devices, so I'm filing this as an issue rather than an untested PR — the correct reserved address is a layout decision best made with hardware to validate. Happy to help implement/test once a safe sector is chosen.

Fix: https://github.com/libretiny-eu/libretiny/pull/408 adds the missing partition entry. (An earlier ESPHome-side runtime guard, esphome/esphome#18647, was closed: @Dr-Blank confirmed the erase happens during SDK bring-up before ESPHome code runs, so a runtime guard is structurally too late — the partition table is the only fix.)

Bl00d-B0b:
Fixes #407. Downstream: esphome/esphome#18646.

## Problem

The Beken BDK BLE bond store (`driver/ble/ble_5_x/.../app_sec.c`) erases the flash sector at `BK_PARTITION_BLE_BONDING_FLASH`'s start address during SDK BLE bring-up (`func_init_extended`/`app_pre_start`), before any application code runs, whenever BLE is compiled in (`CFG_SUPPORT_BLE=1`).

The beken-72xx partition table (`cores/beken-72xx/base/wraps/BkDriverFlash.c`) did not define `[BK_PARTITION_BLE_BONDING_FLASH]`. As a designated-initialiser array it left that slot zero, so `bk_flash_get_info()` returned a pointer to a zeroed entry — `partition_start_addr == 0` — and the BDK erased **flash sector 0, the bootloader**, permanently bricking the device (also over OTA; safe mode cannot run without a bootloader).

Confirmed on hardware by @Dr-Blank (BK7238 / Tuya T1-3S): it bricks even with `enable_on_boot: false` and no tracker, i.e. purely from BLE being linked in — an ESPHome-side runtime guard was proven too late.

## Fix

Each beken flash layout declares a `ble_bonding` partition on a 4 KB sector that is free on that specific layout (no single arithmetic default is safe on all of them):

| layout | sector | sits between |
|---|---|---|
| `beken-72xx` | `0x1E1000` | calibration / userdata |
| `beken-7231n-tuya` | `0x1D1000` | calibration / kvs |
| `beken-7238-tuya` | `0x1E2000` | diff2ya / calibration |
| `beken-7252` | `0x3FF000` | tail sector after calibration |

Sector provenance: on the 7231T/7231N layouts the claimed sectors are the former `net`/`tlv` partitions removed in 9598759 (nothing in-tree has read them since; `BK_PARTITION_NET_PARAM` moved to a hardcoded `0x1FF000`) — a future restore of `net`/`tlv` at their stock addresses would collide with `ble_bonding`, which is why the placement is also recorded in a comment at the partition entry.

The offset/length flow through the standard `FLASH_BLE_BONDING_OFFSET`/`_LENGTH` macros into the partition table, FAL and the generated docs. The placement invariants (4 KB sector alignment, minimum one sector, no overlap with any other partition — including relocated ones via `custom_flash`) are enforced at build time in `builder/utils/flash.py`, and a layout that reaches the partition table without a `ble_bonding` partition hits an explicit `#error` instead of bricking at runtime. The wrapped flash API is hardened as well: `bk_flash_get_info()` reports unmapped enum values as not-found instead of handing out a zeroed entry, and read/write/erase reject out-of-partition ranges.

`BK_PARTITION_BLE_BONDING_FLASH` only exists since BDK **3.0.76** (the default pin is 3.0.33, which has no BLE bond store and no boot-time erase), so the partition entry and guard are gated on `CFG_BDK_VERSION >= 30076`. Verified both ways on `cb2s` (BK7231N): the default 3.0.33 builds clean with the entry compiled out, and 3.0.78 compiles the entry and guard in. (BK7231Q can't serve as the cross-check — `SOC_BDK[SOC_BK7231] = None`, no BDK version builds it.)

Side effect worth noting: on `beken-7252` the new tail sector raises the derived `FLASH_LENGTH` from `0x3FF000` to `0x400000` (matching the board's 4 MiB `upload.flash_size`), which propagates to the FAL `root` partition and `lt_flash_get_size()`.

No real bonding occurs for a scan-only BLE proxy, so the sector is only erased once at init and left blank.

## Testing

Validated on hardware by @Dr-Blank on two layouts:

- **BK7238 (Tuya T1-3S, `0x1E2000`)** — the board that bricked on stock v1.13.0: bootloader byte-identical across boots with BLE active (was 3980/4096 bytes erased), stack up, WiFi connects, sustained advertisement reception, and OTA-after-BLE clean.
- **BK7231N (CB3S, `0x1D1000`)** — BLE fully functional; a full 2 MB before/after flash diff shows only `app` and `kvs` changed (bootloader, `download`, `calibration`, `userdata` byte-identical), and a test pattern written to the bond sector survives BLE bring-up and power cycles — the placement is confirmed harmless (the 7231N BDK defers bond-store writes until an actual bond). OTA-after-BLE crosses the `download` region where the old arithmetic default would have landed, and completes cleanly.

The remaining layouts' sectors are verified free against their JSONs; the `_base` merge order resolves each board to its intended offset (the `beken-7231t-tuya` boards inherit `0x1E1000`). Not hardware-covered: the beken-72xx/7252 layouts, the `CFG_BDK_VERSION < 30076` compile-out path, and an actual bonding write — all inert for the scan-only proxy use case.

To build ESPHome against this branch, override the platform in your device YAML:

```yaml
platformio_options:
  platform: https://github.com/Bl00d-B0b/libretiny.git#fix/bk72xx-ble-bonding-partition
```


Dr-Blank:
**works. Bootloader intact, device boots, BLE stack comes up, WiFi connects.** Full results below

Built exactly as you suggested, no ESPHome external component:

```yaml
platformio_options:
  platform: https://github.com/Bl00d-B0b/libretiny.git#fix/bk72xx-ble-bonding-partition
```

Otherwise identical to the build that bricked: same `bk72xx_ble_tracker`, same scan parameters (`active: false`, `interval: 100ms`, `window: 30ms`), same board, same everything. Flash 83.2% (900,968 / 1,083,136), RAM 23.1%.

The device confirms the branch on its own, so this is not a stale-platform result:

```
[C][libretiny:012]: LibreTiny:
  Version: v1.13.0+sha.f09431d on t1-3s, compiled at Aug 22 2026 18:18:27
```

`f09431d` is the head of the PR branch.

## Flash results

```
$ cmp -l p408-bl-before.bin   p408-bl-after.bin   | wc -l
0
$ cmp -l p408-bond-before.bin p408-bond-after.bin | wc -l
0
$ sha256sum p408-bl-*.bin p408-bond-*.bin
e3c9784f5b5ba269523837505aa17e4178885d68061d2f1aedc91a2ff58708c3  p408-bl-after.bin
e3c9784f5b5ba269523837505aa17e4178885d68061d2f1aedc91a2ff58708c3  p408-bl-before.bin
f47a8ec3e9aff2318d896942282ad4fe37d6391c82914f54a5da8a37de1300c6  p408-bond-after.bin
f47a8ec3e9aff2318d896942282ad4fe37d6391c82914f54a5da8a37de1300c6  p408-bond-before.bin
```

**Bootloader: 0 differing bytes** (was 3980 on every previous BLE build), and `e3c9784f...` is the known-good original hash — not the post-erase `55ae2b72...`.

### On the bonding sector's zero diff


`f47a8ec3e9aff2318d896942282ad4fe37d6391c82914f54a5da8a37de1300c6` is the sha256 of 4096 × `0xFF`:

```
>>> hashlib.sha256(b'\xff'*4096).hexdigest()
'f47a8ec3e9aff2318d896942282ad4fe37d6391c82914f54a5da8a37de1300c6'
```

And BLE demonstrably did initialise, so the "never ran" branch is ruled out directly:

```
[D][bk72xx_ble:189]: BLE stack initialised
[gapm_cmp_evt_handler] conidx:0,operation:0x61,status:0x0
[gapm_cmp_evt_handler] conidx:0,operation:0x64,status:0x0
[D][bk72xx_ble_tracker:319]: Scan started (passive, window=30ms, interval=100ms)
[C][bk72xx_ble:245]: BK72xx BLE:
  MAC address: <ble-mac>
  Active: YES
```

The BDK bond store ran, read a valid non-zero `partition_start_addr`, and left sector 0 alone.

## Boot

Gets past the point it used to die at, with no reset loop:

```
[I][app:151]: ESPHome version 2026.8.0 compiled on 2026-08-22 18:18:02
...
D [   2.178] WIFI: Start OK
...
wpa S: GROUP_HANDSHAKE -> COMPLETED
[V][wifi_lt:461]: Connected ssid='<ssid>' channel=8, authmode=WPA2 PSK
[I][wifi:1609]: Connected
[C][wifi:1268]:   IP Address: <device-ip>
[D][api:255]: Accept <ha-ip>
[D][api.connection:2593]: Home Assistant 2026.8.2 (<ha-ip>): connected
[V][status_led:014]: New app state 0x00
```

Status LED goes out, Home Assistant attaches, the web server is reachable. I power-cycled and captured a second full boot — identical, still healthy, bootloader still `e3c9784f...` afterwards.

For comparison, every previous BLE build on this board produced `WIFI: Initializing func&app` repeating and then permanent silence.


from my side this fixes it. Happy to retest anything after review changes.

Dr-Blank:
Following up on the open item from my last comment — here is the positive proof of received advertisements. **BLE reception works.** Same branch, same board, bootloader still intact.

## What I changed

Nothing about the platform; only the ESPHome side, to make reception observable. The earlier build had no BLE listener configured, which turns out to make a working scanner and a dead one look identical:

- `bk72xx_ble_tracker` emits no per-device log line under continuous scanning — the "Found device" line in `ble_device_base` is gated on one-shot scans.
- Advertisements are not parsed at all unless something listens: `AdvDispatcher::dispatch()` guards the parse path with `#ifdef ESPHOME_BLE_DEVICE_BASE_LISTENER_COUNT`.

So this build adds a `ble_scanner` text sensor (the listener) and `bluetooth_proxy`.

## Result

```
[C][bk72xx_ble_tracker:201]: BK72xx BLE Tracker:
  Scan Duration: 300 s
  Scan Interval: 100 ms (160 BLE units)
  Scan Window: 30 ms (48 BLE units)
  Scan Type: PASSIVE (configured PASSIVE)
  Continuous Scanning: YES
[C][bk72xx_ble:245]: BK72xx BLE:
  Active: YES
[C][bluetooth_proxy:180]: Bluetooth Proxy:
  Mode: advertisement-only (no GATT connections)
  Configured scan: passive
[C][libretiny:012]: LibreTiny:
  Version: v1.13.0+sha.f09431d on t1-3s
```

Advertisements, from the first two seconds of uptime:

```
[V][text_sensor:037]: 'Last BLE Device': {"timestamp":12,"address":"<mac-1>","rssi":-56,"name":""}
[V][text_sensor:037]: 'Last BLE Device': {"timestamp":12,"address":"<mac-2>","rssi":-67,"name":""}
[V][text_sensor:037]: 'Last BLE Device': {"timestamp":12,"address":"<mac-3>","rssi":-61,"name":""}
[V][text_sensor:037]: 'Last BLE Device': {"timestamp":12,"address":"<mac-4>","rssi":-65,"name":""}
[V][text_sensor:037]: 'Last BLE Device': {"timestamp":13,"address":"<mac-5>","rssi":-58,"name":""}
[V][text_sensor:037]: 'Last BLE Device': {"timestamp":14,"address":"<mac-1>","rssi":-58,"name":""}
[V][text_sensor:037]: 'Last BLE Device': {"timestamp":14,"address":"<mac-2>","rssi":-72,"name":""}
```

(MACs redacted)

Five distinct addresses within the first two seconds, a mix of public and random-static, plausible and varying RSSI, sustained continuously afterwards. Empty `name` fields are expected for a passive scan; names ride in scan responses, which need `active: true`.

The devices also appear in Home Assistant's Bluetooth integration through the proxy.

## Summary for this PR

| | result |
|---|---|
| Bootloader `0x0-0x11000` after boot with BLE compiled in | **0 bytes changed** (was 3980) |
| Bonding sector `0x1E2000` | blank before and after — sha256 matches 4096 x `0xFF`, so the erase is a byte-wise no-op on a correctly chosen free sector |
| BLE stack | initialises, `Active: YES` |
| Scanning | receives advertisements continuously |
| Home Assistant | devices visible via `bluetooth_proxy` |

Both halves confirmed: **bootloader intact and BLE actually working.**

Happy to retest after any review changes.

bdraco:
@esphbot review

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Default bonding offset lands inside a declared partition on 3 of the 4 beken flash layouts

Bl00d-B0b:
Thanks — that's a critical catch, the arithmetic default was only safe on the one layout I could validate. Reworked in e5c77a9:

**Per-layout partitions.** Each BLE-capable base layout now declares a `ble_bonding` flash partition at a sector free on *that* layout, so the builder emits `FLASH_BLE_BONDING_OFFSET`/`_LENGTH` and the sector is visible to FAL, the docs and ltchiptool (your suggested approach):

| layout | offset | free sector |
|---|---|---|
| `beken-72xx` | 0x1E1000 | gap calibration→userdata |
| `beken-7231n-tuya` | 0x1D1000 | gap calibration→kvs |
| `beken-7238-tuya` | 0x1E2000 | gap diff2ya→calibration (hardware-validated) |
| `beken-7252` | 0x3FF000 | tail gap above calibration |

I used the gaps I could confirm from the JSONs — same candidates you identified. **These still want board-owner confirmation against real dumps** for every layout except `beken-7238-tuya` (that one @Dr-Blank has validated end-to-end). `beken-7231t-tuya` inherits the `beken-72xx` base entry (0x1E1000) through the layout merge; if that's wrong for a real 7231t board its owner can add an explicit entry.

**Build guard.** Replaced the arithmetic fallback with a compile-time negative-array-size check (no `_Static_assert` under -std=gnu99) that fails the build if the offset is 0 or not 4 KB aligned — a bad override now breaks the build instead of a device.

**Comment fixed** on all three points: erased every boot (not once), a bonding peripheral *writes* here (so overlap is actively harmful, not a no-op), and per-layout placement rather than "the Tuya layouts".

The one thing I can't do from here is validate the non-7238 sectors on hardware. If any board owner can dump `0x0-0x11000` before/after a BLE boot on their layout, that closes it; otherwise the guard at least guarantees no silent brick.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
All six notes addressed, in f80a9d7:

1. **`FLASH_BLE_BONDING_LENGTH`** — the table entry now uses the macro; the dead `#ifndef` fallback is removed (flash.py always emits OFFSET and LENGTH as a pair, and the missing-layout case now fails at the `#error` below).
2. **Missing-macro diagnostic** — `#else` / `#error "Board flash layout must declare a 'ble_bonding' partition (see boards/_base/beken-*.json)"`.
3. **Guard coverage** — the guard now requires the offset to sit at/above the end of the app image (`>= FLASH_APP_OFFSET + FLASH_APP_LENGTH`), which subsumes the old `!= 0` term and closes the inside-`app`/inside-bootloader class.
4. **BDK 3.0.33 / BK7231Q** — your "unverified" item was a live blocker, thanks for insisting on it. `BK_PARTITION_BLE_BONDING_FLASH` does not exist before BDK **3.0.76** (checked every tag: 3.0.33–3.0.70 lack it), so the unconditional initializer broke every default-BDK build. The entry and guard are now gated on `CFG_BDK_VERSION >= 30076` — semantically right too, since older BDKs have no bond store and no boot-time erase. Verified both ways on cb2s/BK7231N: default 3.0.33 builds clean (entry compiled out), 3.0.78 compiles entry + guard in. A wa2 build can't serve as the cross-check — `SOC_BDK[SOC_BK7231] = None`, no BDK version supports BK7231Q at all.
5. **JSON address order** — `ble_bonding` moved to its address-ordered position in all three affected layouts (7252's tail placement was already ordered).
6. **BK7252 `FLASH_LENGTH`** — noted in the description (0x3FF000 → 0x400000, matching the board's 4 MiB `upload.flash_size`).

PR description updated to describe the per-layout design, the BDK gate, and Dr-Blank's hardware validation.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
Both notes addressed, in 6274737:

1. **Provenance** — recorded in both places: a comment at the partition entry (former `net`/`tlv` sectors removed in 9598759, NET_PARAM's move to hardcoded 0x1FF000, and the collision risk if they are ever restored at stock addresses) and a matching line in the PR description.
2. **Guard typedef** — renamed to `ble_bonding_offset_must_be_sector_aligned_and_above_app` so the negative-array diagnostic states the invariant, and the comment now gives the real rationale (portability choice, not a gnu99 `_Static_assert` limitation — thanks for the correction).

The merged-FAL key-order note is understood as pre-existing and name-resolved; left as is.

Re-verified both ways after the change: cb2s on the default BDK 3.0.33 (entry compiled out) and on 3.0.78 (entry + guard in) both build clean.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
Length term added to the guard in 2860736 (`FLASH_BLE_BONDING_LENGTH >= 0x1000`), so a custom layout whose gap-free recompute collapses the bond store to zero bytes fails the build instead of emitting a silently wrong table entry. cb2s/BDK 3.0.78 re-verified clean.

The pre-existing NET_PARAM 0x1FF000 observation is a fair catch for a follow-up — kept out of this PR's scope.

Dr-Blank:
Tested 2860736 on two boards. ESPHome 2026.8.0, `LT_VERSION=1.13.0+sha.2860736`, `CFG_BDK_VERSION=30078` on both.

**T1-3S / BK7238** (`beken-7238-tuya`, `ble_bonding` at `0x1E2000`) — works. Boots, WiFi connects, BLE stack initialises, advertisements received continuously. This is the board that bricked on stock 1.13.0 (bootloader erased, 3980/4096 bytes to 0xFF, every boot), so the fix does what it says.

**CB3S / BK7231N** (`beken-7231n-tuya`, `ble_bonding` at `0x1D1000`) — first hardware for this layout, as far as I know.

BLE is fully functional: continuous passive scan, `Active: YES`, advertisements received at a steady rate.

No collateral damage. I dumped the full 2 MB over serial before flashing and again after BLE had been running across several power cycles, and diffed per-partition. bootloader, `download`, `calibration` and `userdata` were byte-identical throughout; only `app` (the new firmware) and `kvs` (ESPHome preferences) changed.

One thing worth flagging: **the bond sector is never touched on this chip.** I wrote a known 4 KB pattern into `0x1D1000`, booted with BLE scanning confirmed active, power-cycled several times, and the pattern was still there byte-for-byte. A control pattern in the adjacent free sector also survived. So on BK7231N the BDK bond store does not appear to erase at bring-up the way it does on BK7238, with the same BDK version and the same scan-only config.

That means this run confirms `0x1D1000` is **harmless** on real BK7231N hardware, but it does not exercise it — a wrong address there would look identical to a right one. The placement still rests on the layout table rather than on observation.

**OTA after BLE** — both boards took an OTA update after BLE had been running, and came back up with BLE still scanning. On BK7231N that specifically exercises `download` (`0x12A000–0x1D0000`), which is where the earlier arithmetic default (`0x1CF000`) would have landed; a corrupted staging region would surface as a failed OTA rather than a crash.

Not covered: the `beken-72xx` and `beken-7252` layouts (no hardware), the `CFG_BDK_VERSION < 30076` compile-out path, and actual bonding — a scan-only proxy never writes a bond record, so this tests the erase behaviour only.

Bl00d-B0b:
Thank you — that is a thorough validation, and the CB3S run is the first hardware data point for the BK7231N layout.

Folding your matrix into the PR description so the review has it in one place:

- **BK7238 (T1-3S)** — the board that bricked on stock 1.13.0: bootloader byte-identical across boots with BLE active, stack up, continuous reception, OTA-after-BLE OK.
- **BK7231N (CB3S)** — BLE fully functional; full 2 MB before/after diff shows only `app` and `kvs` changed; the bond sector at `0x1D1000` provably untouched (pattern survives power cycles), so the placement is confirmed harmless there even though bring-up never exercises it on that chip; OTA-after-BLE specifically crosses the `download` region where the old arithmetic default would have landed.

Your flag about the bond sector never being touched on BK7231N matches the BDK behaviour difference we saw in bench traces (the 7238 bond store erases at bring-up; 7231N defers until a bond write). Since a scan-only config never writes a bond record on either chip, the erase path is the one that mattered for the brick, and it is the one your 7238 run exercises directly.

Still open, as you note: the beken-72xx/7252 layouts (no known hardware in this thread), the `CFG_BDK_VERSION < 30076` compile-out path (compile-tested only), and a real bonding write. All three are inert for the scan-only proxy use case this PR unblocks.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
Both addressed in 0a27a8b.

**1. `flash.json` label** — `"ble_bonding": "BLE Bonding"` added next to the other beken-specific keys, so regenerated board docs name the sector instead of rendering an "Unknown" 4 KB hole.

**2. Whole-table overlap validation** — added in `env_add_flash_layout()`, exactly where the full layout is in scope: partitions are sorted by offset and any pair where `end(a) > start(b)` fails the build with a readable error naming both partitions and their ranges. It catches the case the C guard cannot see — your `ble_bonding: 0x150000` example lands inside `download` and now fails as `Flash layout error: partition 'download' (0x132000+0xA6000) overlaps 'ble_bonding' (starting at 0x150000)` — and it validates every family's layout, custom offsets included, not just beken.

I kept the C-side guard rather than collapsing it to the bare `#ifdef`: its app-end term is a beken-specific invariant (the bond store must sit above the app image) that the generic generator check deliberately doesn't know about, and the alignment/length terms cost nothing as defense-in-depth behind the new structural check. If you'd prefer the collapse anyway, it's a two-line change.

Verified: the overlap check flags the review's example and passes every in-tree layout; `flash.json` parses; black clean.

Bl00d-B0b:
Correction to my previous comment, worth being explicit about: the first push of this round carried a whole-table overlap check, and validating it against every in-tree board before settling showed that would have been wrong — **48 layouts overlap by design** (`tuya` storage lives inside `userdata` on every Tuya board, and `rdp` inside `tuya` on the wr-series). Aliased partitions are a deliberate feature of the layout model, not an error.

The check now in a79241d is scoped to the actual invariant: **`ble_bonding` is the one partition the SDK erases unilaterally at bring-up, so it alone must not share space with anything.** It is validated against the full table in `env_add_flash_layout()` (custom offsets included), with the comment explaining why the scope is deliberate. A misplaced custom offset still fails with a readable error naming both partitions:

```
Flash layout error: 'ble_bonding' (0x150000-0x151000) overlaps 'download' (0x132000+0xA6000); the SDK erases this sector at BLE bring-up, so it must not share space with any other partition
```

Re-verified against all merged in-tree board layouts: 29 boards carry `ble_bonding`, zero overlaps flagged, and the 48 intentional aliases are untouched. The `flash.json` label from the previous push is unchanged.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
Closed in 6f9d554, by your first option — the ordering analysis was exactly right.

**Custom offsets are now validated in `env_parse_custom_flash_layout()`, against the pre-recompute lengths**, where the original extents still exist:

- a requested `ble_bonding` offset that lands strictly inside an un-overridden partition's original range fails with an error naming the partition and explaining that the recompute would otherwise silently shrink it — your `0x150000`-inside-`download` walk now rejects instead of truncating `download` to `0x1E000`;
- the recompute itself now rejects any partition collapsed to zero length (the equal-offset case — `ble_bonding` on top of `calibration` — fails with a named error instead of letting the SDK erase the RF calibration sector);
- partitions the user is deliberately re-placing in the same `flash:` block stay exempt, so intentional layout surgery keeps working — verified: a legitimate move of `ble_bonding` into genuinely free space is still accepted.

The `AddFlashLayout` comment no longer claims custom coverage: it now states the two-layer split explicitly (board-JSON collisions caught there, custom offsets caught pre-recompute) so a future maintainer doesn't rely on coverage that lives in the other function.

Verified: both of your walk-through cases reject with readable errors, the legitimate-move case passes, all 29 in-tree `ble_bonding` layouts still validate clean, black clean.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
Both closed in 444cf63.

**1. Deduplicated** — your two helpers, verbatim shape: `_parse_range()` (used by both validators; the emit loop keeps its offset/length split since it needs both halves for the defines) and `_find_bond_conflict(flash_layout, start, end, skip)` carrying the ble_bonding-only policy and its rationale in one docstring. The custom-path call is `(requested, requested + 1, skip=set(opts))`, the board-JSON call is the merged bond range with no exemptions — the one real difference is now a call-site argument. A side effect worth noting: the equal-offset case is now caught by the earlier, clearer "lies inside" error (the half-open interval touches the partition start), with the zero-length guard remaining as the net for same-offset collisions among simultaneously re-placed partitions.

**2. Split diagnostics** — `length < 0` now reports the real cause: "Custom offset 0x300000 for partition 'ble_bonding' lies past the end of flash (0x200000)"; the shared-offset message stays for `== 0` only.

Verified by executing the module's real code against the scenario matrix: inside-`download` rejected naming download, equal-to-`calibration` rejected naming calibration, past-end rejected with the flash size, a legitimate move into free space accepted, a board-JSON base-override collision rejected at the second layer, stock layouts clean. black clean.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
Closed in the current head — your two lines, placed beside the overlap check with a comment naming why they exist there (the C guard's alignment/min-size terms are gated on `CFG_BDK_VERSION >= 3.0.76`, so with the default 3.0.33 pin a bad layout would otherwise build green for everyone who doesn't pin a newer BDK).

Verified by executing the module's real code: a misaligned `ble_bonding` (`0x1E1800`) and an undersized one (`+0x800`) both reject with the new message; all 29 in-tree bonding layouts still pass unchanged — a regression guard, as you framed it. black clean.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
All of it closed in ee6faf7 — the three checklist items and the three silent-failure guards.

**1 / 4 (HIGH): custom-offset skip hole** — deliberately moved partitions are now *translated*, not skipped: each is checked at its requested offset with its original length, so a bond store landing inside another moved partition fails too. Your `ble_bonding=0x50000` + `app=0x11000` walk now rejects naming `app`.

**2: C guard demoted honestly** — kept as a last-resort duplicate (the `#ifdef`/`#error` half still carries unique value), with the comment rewritten to say the invariants are enforced in `builder/utils/flash.py` for every build and BDK, and that the "above app end" term is self-satisfying on the custom path.

**3: boundary diagnostics** — `length == 0` splits on whether a next partition exists: shared-offset keeps its message; the last-partition-at-end-of-flash case now names the real cause.

**5 (HIGH): `bk_flash_get_info` returns NULL for unpopulated slots** — a zero-filled designated-initializer slot (`description == NULL || length == 0`) reports not-found instead of handing out address 0. This retires the bug class for *every* unmapped `BK_PARTITION_*` id and covers the uninspectable 3.0.56-3.0.75 window.

**6 (HIGH): erase wrap mirrors the write path** — NULL check (`kNotFoundErr`), overflow-safe bounds check (`kParamErr` when the range exceeds `partition_length`), before touching the flash device. An erase on an unmapped or over-long range can no longer wipe neighbouring sectors while reporting success.

**7: no silent compile-out** — `#ifndef CFG_BDK_VERSION` → `#error` (closing the `-Wno-undef` hole), and a `#pragma message` when a **BLE-enabled** build (`CFG_SUPPORT_BLE`) runs on a BDK below 3.0.76, so relying on the runtime guards alone is visible in the build log without spamming non-BLE builds. The threshold itself stays at 3.0.76 — the enum doesn't exist below it, so the mapping cannot; the runtime guards from 5/6 are the cross-version protection.

**8: beken layout without `ble_bonding` fails in Python** — for every build, not just the version-gated C `#error`: an `mcu` starting with `bk` whose merged layout lacks the key raises with a pointer at the base JSONs. Family-scoped, so RTL/LN layouts are untouched.

Verified by executing the real module: ten scenarios (both original brick cases, past-end, at-end boundary, legit single and double moves, bond-inside-moved-partition, beken-without-bonding, non-beken-without-bonding) all behave with cause-naming errors, and **all 63 in-tree board layouts pass the full validation unchanged**. black clean.

Bl00d-B0b:
Simplified per re-evaluation: dropped the redundant guards (C negative-array typedef, CFG_BDK_VERSION presence check, pragma message, Python beken-presence check - old BDKs have no bond store and newer ones trip the #error) and trimmed the comments down to the essentials. Behavior unchanged: same validators in flash.py, same runtime NULL/bounds guards, same #error for a missing ble_bonding partition. Re-verified the full scenario matrix and all 63 board layouts. Net diff is now +131/-6.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
Both suggestions addressed in 21a6c2e:

1. NET_PARAM: bk_flash_get_info() now also returns NULL for zero-length entries, so read/write/erase agree (all kNotFoundErr). Nothing in-tree reads NET_PARAM through the wrap, and its hardcoded 0x1FF000 lies inside userdata on every layout, so blocking the stray write is the right side to pick.

2. Custom-path effective table now caps each partition's end at the next requested offset, so vacated space is no longer attributed to the original occupant. One deviation from the suggested snippet: ble_bonding is excluded from the ordered list entirely, not just from emission - if the bond's own offset participates in the capping, a bond placed inside an unmoved partition caps that partition at the bond offset and the conflict check never fires (e.g. ble_bonding=0x100000 inside app would pass). Verified: calibration=0x1D0000 + ble_bonding=0x1D2000 on beken-72xx is now accepted, the relocated-partition case names the actual occupant, and the bond-inside-app case still fails.

Re-ran the full scenario matrix (16 cases) and all 63 board layouts.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
Addressed in 5d14a1a:

- Sector provenance is now recorded in the comment at the BK_PARTITION_BLE_BONDING_FLASH entry (one line per layout: 0x1E1000 and 0x1D1000 are the former net/tlv sectors from 9598759, 0x1E2000 is free between diff2ya and calibration, 0x3FF000 is the 7252 tail sector), and the stale 'compile-time guard' sentence in the description now describes the actual enforcement in builder/utils/flash.py.
- Took the read/write alternative for NET_PARAM instead of the NULL return: get_info again tests only the description, so BDK callers never see NULL for a populated slot, and bk_flash_read/bk_flash_write now bounds-check the range like erase does. Net effect for the zero-length NET_PARAM entry is the same (any non-empty read/write/erase fails with kParamErr) without widening the get_info contract.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
All three items from the last round, in 40cc501:

1. Extracted _validate_custom_bond_offset() (pre-recompute path) and _validate_bond_placement() (board-JSON path) next to _find_bond_conflict(), which now takes exclude= instead of hardcoding the name. Both call sites are one line and the recompute loop is readable on its own again.

2. All four new failure paths now use click.secho(fg="red") + exit(1) via a small _fatal() helper, matching builder/family/beken-72xx.py. The 'Unrecognized template key' ValueError stays as-is per your note (internal bug, not user config).

3. MAC residual - closed with binary evidence instead of a hardware run: WIFI_MAC_POS is MAC_RF_OTP_FLASH, so cfg_load_mac() goes through manual_cal_get_macaddr_from_flash() in libcal_bk7231n.a, and disassembling that archive shows all 18 bk_flash_get_info call sites load r0 = #3 (BK_PARTITION_RF_FIRMWARE) before the call. Nothing in the cal library (or anywhere else linked) touches BK_PARTITION_NET_PARAM, so the station MAC never flows through the entry this PR changes. The RF_FIRMWARE reads land inside its real 0x1000-length entry and pass the new bounds checks.

Re-ran the 16-case scenario matrix (messages byte-identical, now via the red-error path) and all 63 board layouts; black and clang-format clean.

esphbot (approved):
> [!TIP]

Bl00d-B0b:
8f21e5d: `_fatal` annotated `-> NoReturn` (with the `typing` import); matrix and 63-board validation re-run, black clean.

esphbot (approved):
> [!TIP]

PR #410, opened 2026-08-23
[ambd] Add BLE scan support (Realtek GAP)

Bl00d-B0b:
Enables compilation and linking of the Realtek AmebaD BLE stack, on top of #403 — LibreTiny provides no BLE API of its own.

Per the maintainer direction below, this PR is build enablement only, the beken-72xx arrangement: the family builder compiles the SDK's HCI/coexistence glue and links `btgap.a`, gated on `CONFIG_BT=1` in the `platform_opts_bt.h` custom options (the same `env.Cfg` mechanism beken uses for `CFG_SUPPORT_BLE`). Without the option, nothing BT-related is compiled or linked.

The vendor-SDK invocation (GAP bring-up, message pump task, scanning, WiFi/BT coexistence sequencing) lives in the consumer — ESPHome's `rtl87xx_ble` component ([esphome/esphome#18720](https://github.com/esphome/esphome/pull/18720)) — mirroring how ESPHome's `bk72xx_ble` drives the Beken BDK directly.

## Testing

Verified on BW16 (RTL8720DN) hardware through the ESPHome consumer: stack bring-up after WiFi STA with coexistence enabled, continuous advertisement scanning with WiFi connected, active scanning with scan-response delivery, and repeated OTA cycles. A build without the option compiles nothing BT-related.


Bl00d-B0b:
API extension in 125a888, driven by downstream review ([esphome/esphome#18720](https://github.com/esphome/esphome/pull/18720)):

- `lt_ble_adv_cb_t` now carries the GAP advertising event type (`T_LE_SCAN_INFO.adv_type`, Core spec values) so the consumer can demux advertisements from scan responses and merge the pair.
- `lt_ble_scan_running()` now reads the GAP stack's own device state (`GAP_PARAM_DEV_STATE.gap_scan_state`) instead of a local flag, so a scan dropped inside the controller (coexistence pressure) is visible to the caller's recovery logic.

Hardware-verified on BW16: with an active scan, scan-response-only device names now appear in merged frames (59/106 named vs 11/92 passive).

kuba2k2:
Hi,

I currently don't have the time to review all these PRs, but I'm noticing something concerning that I should probably point out before it's too late:

- LibreTiny doesn't have BLE support, period. No other family (bk72xx, rtl87xx) has BLE APIs ***within LibreTiny***. The simplest reason is that all provided APIs should be identical on all supported families - that's the core goal LibreTiny is trying to achieve. Adding a BLE API to RTL8720D just won't work here.

If BLE support in ESPHome is requested already, then it's probably best for the vendor SDK invocation to be moved to an ESPHome component - just like it's done for BK72xx (AFAIK). What LT can do is enable compilation/linking of any SDK libraries needed to use BLE - again, just like with BK72xx (there's a build script-level check that links BLE if necessary).

---

Side note, LibreTiny will probably never have its own BLE API - I'm considering implementing (porting) BTstack instead. This would have the obvious benefit of not having to mess with the vendor SDKs (as well as a stable, shared API across all platforms).

But then I'm also considering writing a custom build system... so yeah, there's a lot of plans and a lot of other things to do in life.

Bl00d-B0b:
Thanks — that's a fair line to draw, and it matches how the other chips ended up working, so I've reshaped this PR to exactly the beken-72xx arrangement you describe (in a22a345):

- `lt_ble.c`/`lt_ble.h` are gone — no BLE API in LibreTiny.
- What remains is compile/link enablement only: the family builder compiles the SDK's HCI/coex glue and links `btgap.a`, gated on `CONFIG_BT=1` in the `platform_opts_bt.h` custom options (the same `env.Cfg` mechanism beken uses for `CFG_SUPPORT_BLE`). Without the option, nothing BT-related is compiled or linked.
- The vendor-SDK invocation (GAP bring-up, message pump, scanning) moved to an ESPHome component — [`rtl87xx_ble` in esphome/esphome#18720](https://github.com/esphome/esphome/pull/18720) — which requests the option, mirroring how ESPHome's `bk72xx_ble` drives the Beken BDK directly.

For context on the ESPHome side: BLE observer support is already in ESPHome for BK72xx and LN882H on a platform-neutral layer (the chip components only feed advertisements into it), and the same arrangement is running on RTL8720D hardware against this branch — advertisement proxy, sensors, active scanning. So nothing here needs LibreTiny to own a BLE API; SDK enablement is all that's asked of the framework.

On the BTstack idea: ESPHome's Pico W support already drives BTstack today, so if LibreTiny grows a BTstack port one day there's a ready consumer pattern for it — the vendor-SDK path here doesn't get in its way and can be retired chip by chip whenever that lands.

If this slimmed scope matches what you had in mind, the PR is ready as-is; happy to adjust further if not.

bdraco:
side note: BTStack requires a license, every rp2040/rp2350 chip comes with one.

bdraco:
We are planning on adding NimBLE support in the future in ESPHome so that might be the direction to go

kuba2k2:
> side note: BTStack requires a license, every rp2040/rp2350 chip comes with one.

Yes - if used commercially. I've contacted the authors some time ago, and they said it wouldn't be a problem to use BTstack in LibreTiny. To best of my knowledge nobody is selling devices based on LT. And even if they did, they could just contact BTstack for a license directly.

Bl00d-B0b:
To make the strategy behind this concrete, the current ESPHome BLE observer/proxy architecture already has this shape:

```
                    Application: sensors / bluetooth_proxy / automations
                                             |
                                   +---------+---------+
                                   |  Shared BLE core  |
                                   | (ble_device_base) |
                                   +---------+---------+
                                             |
                                 +-----------+------------+
                                 |    BLEHub contract     |
                                 | (platform/backend API) |
                                 +-----------+------------+
                                             |
        +--------------+--------------+------+-------+--------------+
        |              |              |              |              |
      ESP32        RP2040+CYW       BK72xx         LN882H        RTL87xx
        |              |              |              |              |
    Bluedroid       BTstack       Beken BDK        LN BLE        Realtek
    / NimBLE*                                      stack        BLE stack
        |              |              |              |              |
   ESP BT ctrl      CYW43439       BK radio       LN radio      RTL radio

  * NimBLE: planned for ESPHome
```

The split matters:

- `ble_device_base` owns Bluetooth behavior shared across platforms: advertisement representation/parsing, scan-response merging, listener dispatch, scanner state, and the consumer-facing BLE semantics.
- The per-platform `BLEHub` implementation absorbs the vendor-specific differences: GAP APIs, task/ISR handoff, stack initialization, coexistence sequencing, retries, and controller quirks.
- Consumers such as sensors, automations and `bluetooth_proxy` only depend on the shared layer / BLEHub contract and do not know which Bluetooth stack is underneath.

Today the backends are therefore intentionally different:

| Chip | Backend today | Possible future direction |
|---|---|---|
| ESP32 | Bluedroid | NimBLE (planned in ESPHome) |
| RP2040 / Pico W | BTstack + CYW43439 | keep BTstack |
| BK7231N / BK7238 | Beken BLE stack | replaceable behind BLEHub if a shared host becomes practical |
| LN882H | Lightning Semi BLE stack | same |
| RTL8720D / RTL8720C | Realtek Ameba BLE stack | same |

nRF52840 is deliberately not shown in the current diagram: ESPHome has Zephyr/nRF52 Bluetooth functionality in other areas, but it does not currently implement this BLEHub observer/scanner path for sensors / `bluetooth_proxy`.

This is why I don't think LibreTiny needs to define a common BLE API itself.

LibreTiny only needs to expose/build the SDK pieces required by a family. The cross-platform BLE API already exists one layer above it in ESPHome through `ble_device_base` / `BLEHub`.

That also leaves the implementation replaceable chip by chip. For example:

```
ESP32: Bluedroid -> NimBLE
```

can happen without changing sensors, automations, bluetooth_proxy, or any other BLE consumer.

The same would be true for BK/LN/RTL if a common host stack such as NimBLE ever becomes technically worthwhile and the controller/HCI boundary makes that practical. There is no need to commit to that today.

For RTL8720D specifically, this PR therefore only needs to enable/link the Realtek Bluetooth SDK pieces required by the ESPHome RTL BLEHub backend (AmebaZ2 needs nothing from LibreTiny at all — its BT stack is already built and linked in released versions, and the ESPHome component now compiles for it in CI). It does not make LibreTiny responsible for a new BLE abstraction, and it does not prevent a different host implementation from replacing the Realtek host later.

PR #414, opened 2026-09-13
[beken-72xx] WiFi sleep: DTIM sleep only, no MCU power save

Bl00d-B0b:
## Problem

`WiFi.setSleep(true)` enables two SDK power save modes: RF DTIM sleep and MCU sleep (`bk_wlan_mcu_ps_mode_enable()`). MCU sleep breaks the chip in two ways:

- **Without BLE:** the CPU halts between wake events, nothing feeds the watchdog, and the chip resets within seconds (`WDT Reset`). Details and measurements in #415.
- **With BLE running:** stopping the station halts the chip for good. Every task stops, UART, GPIO and network are dead, and only a power cycle recovers it. Field reports: esphome/esphome#18592.

Sequence of the BLE halt in the SDK (BDK 3.0.78, `CFG_USE_MCU_PS=1`):

1. `bk_wlan_stop(BK_STATION)` removes the vif and ends with `power_save_rf_hold_bit_clear(RF_HOLD_BY_STA_BIT)`.
2. `rf_hold_status` reaches 0 (BLE only holds the RF per event), so `sctrl_rf_sleep()` powers the RF down.
3. The idle task runs `mcu_power_save()`. `ps_may_sleep()` is true because MCU power save is on, so `sctrl_mcu_sleep()` runs.
4. With the BLE schedule still switching the RF, the MCU never wakes up.

## Change

- `setSleep(true)` enables only the RF DTIM sleep. That is where the idle power saving comes from (measurements in #415).
- `bk_wlan_stop(BK_STATION)` turns DTIM sleep off and nothing turns it back on, so `modePriv()` re-applies the requested sleep mode on the next station start.
- `setSleep(false)` is unchanged and still turns both modes off.

This includes the fix from #415.

## Testing

CB3S (BK7231N), ESPHome 2026.8 with `bk72xx_ble_tracker` scanning continuously, UART1 captured on a bench. A template button runs `wifi.disable`, waits 20 s and runs `wifi.enable`. The station stop goes through `WiFi.mode()`, `modePriv()` and `bk_wlan_stop(BK_STATION)`.

| LibreTiny | WiFi sleep | result |
|---|---|---|
| v1.13.0 | DTIM + MCU | dead after the first press, last UART line is the STA stop event |
| v1.13.0 | off | alive, station re-enabled after 20 s, reconnected (2/2) |
| v1.13.0, MCU sleep kept off while BLE runs | DTIM only | alive, station re-enabled after 20 s, reconnected (2/2) |

The last row is the first revision of this PR. With BLE running it takes the same DTIM-only path as this change.

Last lines on v1.13.0 with sleep on (application log lines removed):

```
WIFI: Disabling STA
hapd remove_if:0
mm_remove_if_req_handler remove vif0
released_scan_results
WIFI: Disconnecting from  (wifiOff=0)
WIFI: BK event 131
```

Nothing follows. The chip does not answer the flasher link until a power cycle.

The watchdog reset without BLE and the power measurements are covered in #415: 15 BK7231T and 2 BK7231N bulbs running DTIM-only sleep.


cores/beken-72xx/arduino/libraries/WiFi/WiFiGeneric.cpp
```diff
@@ -112,7 +129,15 @@ bool WiFiClass::setSleep(bool enable) {
...
-		bk_wlan_mcu_ps_mode_enable();
+		// MCU sleep while the BLE controller runs halts the chip once the station
+		// goes idle: the last WiFi RF hold is dropped, the RF is powered down and
+		// the MCU sleeps under a live BLE schedule; it never wakes up again.
+		// Keep the RF DTIM sleep only in that case.
+		if (&ble_thread_is_up == NULL || !ble_thread_is_up()) {
+			bk_wlan_mcu_ps_mode_enable();
+		} else {
+			bk_wlan_mcu_ps_mode_disable();
+		}
```
ryan-ronnander: MCU sleep also boot loops the chip without BLE (WDT reset within seconds once it's on, see #415), so it can't be enabled at all. DTIM alone is where the idle power saving comes from.


```suggestion
```

cores/beken-72xx/arduino/libraries/WiFi/WiFiGeneric.cpp
```diff
@@ -2,6 +2,12 @@
 
 #include "WiFiPrivate.h"
 
+extern "C" {
+UINT32 mcu_ps_is_on(void);
+// only present when the BLE stack is linked in
+bool ble_thread_is_up(void) __attribute__((weak));
```
ryan-ronnander: Remove: MCU sleep causes a reboot loop. See my other comment for details.

```suggestion
```

ryan-ronnander (commented):
I didn't catch this PR originally, since I had a local power save fix drafted the same day it opened (#415).

The issue is that MCU sleep boot loops the chip, even without BLE (WDT resets once power save is actually on, which ESPHome doesn't do yet until esphome/esphome#19491). These suggestions drop it entirely and keep DTIM, which is where the power savings come from based on my testing.

Once these are in, I'll close #415.

cores/beken-72xx/arduino/libraries/WiFi/WiFiGeneric.cpp
Bl00d-B0b: Applied. `setSleep(true)` now enables only the DTIM sleep. The MCU sleep guard before `bk_wlan_stop(BK_STATION)` is dead code without MCU sleep, so it is gone too. The re-apply on the next station start stays: `bk_wlan_stop(BK_STATION)` turns the DTIM sleep off and nothing in the SDK turns it back on.

cores/beken-72xx/arduino/libraries/WiFi/WiFiGeneric.cpp
Bl00d-B0b: Removed, together with `mcu_ps_is_on()`, whose only user was that guard.

Bl00d-B0b:
Thanks @ryan-ronnander, both suggestions are in. The description now covers the watchdog reset without BLE and points to #415 for your measurements.

PR #415, opened 2026-09-23
[beken-72xx] Fix boot loop when enabling WiFi sleep

### LINKED ISSUE #180: High power consumption on BK7231- power_save_mode in wifi component not working
First of all, I'm really impressed of this unexpected and awesome new feature of esphome thanks to libretiny!
I've uploaded (via UART) my first yaml configuration into single channel relay on CB2S, and it is working from the first try, reliable and stable.
But I've noticed the relay is getting warm idle, so I've measured power consumption. In my case it goes to 1.2 - 1.3 W idle, while same relay witch original tuya-soft consumes 0.5 - 0,6 W.
Apart of unnecessary power consumption (which can be an issue when you count your devices in dozens), the real problem can be the heat. Imagine device closed in wall-socket, without any air flow, dispersing 1.3W of energy in heat 24h/7d - it might lead to serious overheating issues.
I've tried different power_save_mode values of wifi component, usually working good witch esp chips, but witch no effect.
Looks like power_save_mode is not implemented for BK7231 - is it planned to be implemented in future releases?

ryan-ronnander:
> **Superseded by #414**, which includes this change. Closing in favor of it.

One of three fixes that together close #180 (power save not working on BK7231):

1. This PR (libretiny): prevents WiFi sleep from boot looping the chip.
1. libretiny-eu/lwip#5: makes a static IP device report that it has an IP address, so power save can turn on. DHCP already worked.
1. https://github.com/esphome/esphome/pull/19491: turns power save on after the device has an IP address, instead of before WiFi connects, when BK72xx ignores it.

---

## Problem

On BK72xx, `WiFi.setSleep(true)` also calls `bk_wlan_mcu_ps_mode_enable()`. MCU power save halts the CPU between wake events, so the main loop stops, nothing feeds the watchdog, and the chip resets within seconds (`reset_reason`: `WDT Reset`). If the application turns sleep back on after connecting, the device reboots after every connect.

## Fix

This PR makes `setSleep(true)` enable only DTIM radio sleep. That is where the idle power saving comes from: with all three fixes, idle power drops 48% on BK7231T and 38% on BK7231N (table below).

I originally hit this on [SYLVANIA A19CCECWF RGBW bulbs](https://www.elektroda.com/rtvforum/topic4034705.html) running ESPHome with `power_save_mode: light`. The stock Tuya firmware idled lower than ESPHome, which is what sent me looking.

ESPHome users can't hit this boot loop currently, only because ESPHome doesn't actually turn power save on (fix 3 above). 

### Testing

Idle power at the wall, light off:

| Firmware | BK7231T (WB8P) | BK7231N (CBLC5) |
| --- | --- | --- |
| Stock Tuya | [0.43 W](https://github.com/bramstroker/homeassistant-powercalc/blob/v1.25.5/profile_library/sylvania/A19CCECWF/model.json#L31) | [0.41 W](https://github.com/bramstroker/homeassistant-powercalc/blob/v1.25.5/profile_library/sylvania/40A19FILCCLWIFI/model.json#L30) |
| ESPHome, power save off | 0.63 W | 0.65 W |
| ESPHome, power save on (all three fixes) | 0.33 W | 0.40 W |

- Before: my test bulb logged 37 `WDT Reset`s in about 70 minutes with sleep enabled.
- After: no `WDT Reset` over the next two days, until I removed the diagnostic sensor.
- 15 of my BK7231T bulbs have run this change since 2026-09-13 with stable power savings.
- 2 of my BK7231N bulbs have run this change since 2026-09-20 with stable power savings.


ryan-ronnander:
Overlaps with #414, which changes `setSleep()` as well. This PR removes MCU sleep entirely, which also prevents the BLE hang #414 fixes (with mcu_ps_on never set, ps_may_sleep() can't put the MCU to sleep).

#414's table shows power_save_mode: high running fine without BLE on v1.13.0, but that's because ESPHome never actually turns power save on today! (pending fix in https://github.com/esphome/esphome/pull/19491 in the list above): the radio never sleeps, so the MCU doesn't either.

Once power save really turns on, MCU sleep boot loops the chip even without BLE.

Cossid:
Honestly, I'd probably just close this one as duplicate, and update documentation (esphome issue) to rely on #414

ryan-ronnander:
> Honestly, I'd probably just close this one as duplicate, and update documentation (esphome issue) to rely on #414

Here's the thing though, #414 still calls `bk_wlan_mcu_ps_mode_enable()` on devices that don't use BLE, and that call causes the WDT boot loop this PR fixes. So #414 alone leaves every device without BLE broken once esphome/esphome#19491 actually turns power save on.

I missed #414 when I opened this (I had local fixes drafted and running the same day #414 opened). I don't mind which one gets merged. If it's simpler to fold the removal into #414 I'll close this one out.

Cossid:
Ah, I misread what #414 was doing, just noticed the same line removal, but forgot it's replacing it.  I'd say probably suggest an edit to #414 then, as this one will surely end in a merge conflict after #414 gets merged, which seems likely.

ryan-ronnander:
Closing as superseded by #414, which now includes this fix (DTIM-only `setSleep(true)`). Thanks @Bl00d-B0b for folding it in.

